PR #5, opened 2024-03-18
Add Schema::open with simple quickstart_dense test

rroelke:
This pull request adds an equivalent of the `tiledb_array_schema_load` function which opens an array schema from a URI.

Prior work with array schema *creation* put the Domain in as a field in the `ArraySchema`, so most of the implementation here is actually about that.  `ArraySchema::open` calls `Domain::load` which fetches into Rust the information needed from the C API.

"Prior work" was the focus on the `quickstart_dense` example which creates, rather than reads, an array.  In that example we had to make sure that the Domain lifetime was at least as long as the Schema lifetime, and then Domain has a vector of Dimension for the same reason.

Since the tests for this naturally want to ask about the dimensions, there are also changes to the `Domain::dimension` function, specifically to return a reference to an existing one rather than create a new one.  This uses interior mutability to lazily manifest the dimensions from the C API.

### DIFF
--- a/tiledb/api/src/array/dimension.rs
+++ b/tiledb/api/src/array/dimension.rs
@@ -37,12 +37,8 @@ impl<'ctx> Dimension<'ctx> {
     }
 
     /// Read from the C API whatever we need to use this dimension from Rust
-    pub(crate) fn load(
-        context: &'ctx Context,
-        raw: RawDimension,
-    ) -> TileDBResult<Self> {
-        // TODO: ???
-        Ok(Dimension { context, raw })
+    pub(crate) fn new(context: &'ctx Context, raw: RawDimension) -> Self {
+        Dimension { context, raw }
     }
 
     pub fn datatype(&self) -> Datatype {
--- a/tiledb/api/src/array/dimension.rs
+++ b/tiledb/api/src/array/dimension.rs
@@ -8,7 +8,6 @@ use crate::Result as TileDBResult;
 
 pub(crate) enum RawDimension {
     Owned(*mut ffi::tiledb_dimension_t),
-    Borrowed(*mut ffi::tiledb_dimension_t),
 }
 
 impl Deref for RawDimension {
@@ -16,16 +15,14 @@ impl Deref for RawDimension {
     fn deref(&self) -> &Self::Target {
         match *self {
             RawDimension::Owned(ref ffi) => ffi,
-            RawDimension::Borrowed(ref ffi) => ffi,
         }
     }
 }
 
 impl Drop for RawDimension {
     fn drop(&mut self) {
-        if let RawDimension::Owned(ref mut ffi) = *self {
-            unsafe { ffi::tiledb_dimension_free(ffi) }
-        }
+        let RawDimension::Owned(ref mut ffi) = *self;
+        unsafe { ffi::tiledb_dimension_free(ffi) }
     }
 }
 
@@ -37,6 +34,15 @@ pub struct Dimension<'ctx> {
 impl<'ctx> Dimension<'ctx> {
     pub(crate) fn capi(&self) -> *mut ffi::tiledb_dimension_t {
         *self.raw
+    }
+
+    /// Read from the C API whatever we need to use this dimension from Rust
+    pub(crate) fn load(
+        context: &'ctx Context,
+        raw: RawDimension,
+    ) -> TileDBResult<Self> {
+        // TODO: ???
+        Ok(Dimension { context, raw })
     }
 
     pub fn datatype(&self) -> Datatype {
--- a/tiledb/api/src/array/domain.rs
+++ b/tiledb/api/src/array/domain.rs
@@ -1,3 +1,4 @@
+use std::cell::UnsafeCell;
 use std::ops::Deref;
 
 use crate::array::{dimension::RawDimension, Dimension};
@@ -27,12 +28,29 @@ impl Drop for RawDomain {
 pub struct Domain<'ctx> {
     context: &'ctx Context,
     raw: RawDomain,
-    dimensions: Vec<Dimension<'ctx>>,
+
+    /// Array dimensions attached to this domain.
+    /// These must not be freed until the Domain is.
+    /// UnsafeCell is used for interior mutability so that when a dimension
+    /// is queried we can stash it here and
+    dimensions: UnsafeCell<Vec<Option<Dimension<'ctx>>>>,
 }
 
 impl<'ctx> Domain<'ctx> {
     pub(crate) fn capi(&self) -> *mut ffi::tiledb_domain_t {
         *self.raw
+    }
+
+    /// Read from the C API whatever we need to use this domain from Rust
+    pub(crate) fn load(
+        context: &'ctx Context,
+        raw: RawDomain,
+    ) -> TileDBResult<Self> {
+        Ok(Domain {
+            context,
+            raw,
+            dimensions: UnsafeCell::new(vec![]),
+        })
     }
 
     pub fn ndim(&self) -> u32 {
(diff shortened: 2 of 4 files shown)


src/array/dimension.rs
```diff
@@ -39,6 +36,15 @@ impl<'ctx> Dimension<'ctx> {
         *self.raw
     }
 
+    /// Read from the C API whatever we need to use this dimension from Rust
+    pub(crate) fn load(
```
davisp, line 40: I'm not a fan of using `load` here as our code base generally uses `load` to refer to pulling bytes from disk or network.

I'm pretty sure the more idiomatic approach here would be to either use a From trait (or something like it depending on multiple argument support) or a method prefixed with `from_` if multiple arguments means we go the "static method" approach (no clue what Rust calls those).

src/array/domain.rs
```diff
@@ -27,14 +28,31 @@ impl Drop for RawDomain {
 pub struct Domain<'ctx> {
     context: &'ctx Context,
     raw: RawDomain,
-    dimensions: Vec<Dimension<'ctx>>,
+
+    /// Array dimensions attached to this domain.
+    /// These must not be freed until the Domain is.
+    /// UnsafeCell is used for interior mutability so that when a dimension
+    /// is queried we can stash it here and
+    dimensions: UnsafeCell<Vec<Option<Dimension<'ctx>>>>,
```
davisp, line 36: First, the lifetime description is wrong. The pointer you get from core when you call `tiledb_domain_get_dimension_from_index` is to a wrapper handle. It can be deallocated at any time before the owning context is destroyed. The actual internal object is behind a shared_ptr on the pointer we deallocate in Drop traits.

Secondly, I'd probably not attempt to store the dimensions here as a cache. Just have `domain.dimension(index) -> TileDBResult<Dimension<'ctx>>` or w/e.

src/array/domain.rs
```diff
@@ -27,14 +28,31 @@ impl Drop for RawDomain {
...
 }
 
 impl<'ctx> Domain<'ctx> {
     pub(crate) fn capi(&self) -> *mut ffi::tiledb_domain_t {
         *self.raw
     }
 
+    /// Read from the C API whatever we need to use this domain from Rust
```
davisp, line 44: Same not on load -> from

src/array/domain.rs
```diff
@@ -49,24 +67,30 @@ impl<'ctx> Domain<'ctx> {
...
-                context: self.context,
-                raw: RawDimension::Borrowed(c_dimension),
-            })
-        } else {
-            Err(self.context.expect_last_error())
+    pub fn dimension(&self, idx: usize) -> TileDBResult<&Dimension<'ctx>> {
+        let known_dims = unsafe { &mut *self.dimensions.get() };
+        if idx >= known_dims.len() {
+            let mut c_dimension: *mut ffi::tiledb_dimension_t = out_ptr!();
+            let c_idx = idx.try_into().unwrap();
```
davisp, line 74: Pretty sure you want `?` instead of `unwrap()` here.

src/array/domain.rs
davisp: Coming back to re-stress that we should *really* avoid keeping a second state like this. I've never seen it lead anywhere except to subtle bugs when the wrapped copy of state drifts from what the interior state of the domain instance contains.

In these cases, each time we get a dimension, all we're actually getting is a handle class that contains a shared_ptr to the "real" dimension internal to the domain. And those handles are designed to be created cheaply and destroyed frequently *specifically* to reduce the refcounting/lifetime issues in language bindings.

src/array/schema.rs
```diff
@@ -20,41 +21,88 @@ impl ArrayType {
...
+            ffi::tiledb_array_schema_free(ffi)
+        }
     }
 }
 
 pub struct Schema<'ctx> {
     context: &'ctx Context,
     raw: RawSchema,
-    _domain: Domain<'ctx>,
+    domain: Domain<'ctx>,
```
davisp, line 49: Same thing here as for the Domain.dimensions vector. We don't want to hold state here as that means we're doing a shadow state/cache thing which is what fuels my nightmares.

src/array/schema.rs
```diff
@@ -20,41 +21,88 @@ impl ArrayType {
...
 }
 
 impl<'ctx> Schema<'ctx> {
     pub(crate) fn as_mut_ptr(&self) -> *mut ffi::tiledb_array_schema_t {
         *self.raw
     }
 
+    pub fn domain(&self) -> &Domain<'ctx> {
+        &self.domain
```
davisp, line 58: I think this should just be a call to `tiledb_array_schema_get_domain`.

src/array/schema.rs
```diff
@@ -20,41 +21,88 @@ impl ArrayType {
...
+    }
+
+    /// Retrieve the schema of an array from storage
+    pub fn open(context: &'ctx Context, uri: &str) -> TileDBResult<Self> {
```
davisp, line 62: The TileDB API's would call this public API `load` or similar. I would either go with that, or the same thing as for the other `load` calls I mentioned where we might want to be more rust-y and use either a `From` trait or a `from_uri` method.

src/array/schema.rs
```diff
@@ -248,4 +301,40 @@ mod tests {
...
+        let r = create_quickstart_dense(&tmp_dir, &c);
+        assert!(r.is_ok());
+
+        let schema = Schema::open(&c, &r.unwrap())
+            .expect("Could not open quickstart_dense schema");
+
+        let domain = schema.domain();
+
+        let rows = domain.dimension(0).expect("Error reading rows dimension");
+        assert_eq!(Datatype::Int32, rows.datatype());
```
davisp, line 320: Just to illustrate the lifetimes, if you were to put this in a loop so that you called this N times to create the dimension and then drop it, internally it'd be just fine. Shoving the returned dimensions into a vector or similar would also be fine. You'd have N distinct allocated handle objects that all proxy through to the underlying internal `shared_ptr<tiledb::sm::Dimension>`.

davisp (changes requested):
I had a number of comments, the important bits are around lifetime stuff. I know its fairly confusing to intuit some of our APIs given that you've done Domain stuff where the returned pointers are "borrows" that require you to copy stuff out. In these cases where you're holding pointers to internal bits, I'd argue that we should avoid that at all costs as those internal handle classes are designed specifically for allowing wrappers to avoid a lot of complicated lifetime issues.

Or in other words, for almost all of the `tiledb_$thing_t` types, they're meant to be used in an RAII style class. I.e., the pointer gets set in a "constructor" and then freed in the "destructor" where I'm using quotes because there are different names for those in various languages.

An example where we're doing it *almost* correctly is here in the filter list:

https://github.com/TileDB-Inc/tiledb-rs/blob/main/src/filter_list.rs#L69-L90

Almost because that's till using the Borrowed pattern which is a memory leak in these cases. But that approach to just returning an instance of the thing we're getting should be the preference unless there's a very compelling reason why that could be wrong (only thing that comes to mind would be caching expensive calls, but even then, I'd probably suggest that cache be upstream in user code rather than something we do internally. Also, expensive is relative there, as I'd expect performance of these calls to be in the 10-100 kHz range).

src/array/domain.rs
rroelke: I'm in agreement that the less state we keep the better.

I'm going to double check this but the reason that I put the member here in the first place is that I was getting memory errors in the quickstart_dense example with array creation when I first got the API far along enough.  I'm pretty sure approaching it this way fixed it.  I'll check that now because obviously it matters a lot for what the path forward here is.

rroelke: All righty, so... git bisect points at
`58dd0e6dc1ae6b347f050825888d6d75f5083fda is the first bad commit`

where "bad" actually means "does not segfault".

This is the commit where I put the `Vec<Dimension>` in there and so on.

If we look at the C API of `quickstart_dense.c` then we see that the `create_array` function allocates objects, processes them, and then frees them all at the end.  The "good", i.e. faulting, commits do not follow this pattern because they drop the Rust API structs eagerly, e.g.

```
let domain = {
    /* dimensions are created and dropped, and thus freed, within this block, resulting in SEGV */
};
```
whereas moving the dimensions into the `Vec<Dimension>` field of Domain did not drop them until after the domain is done being used.

src/array/dimension.rs
rroelke: I think `new` is also commonly used for this purpose :)

rroelke:
I'm pleased to report that removing the interior secondary state is perfectly fine and I have now done that. Changes are addressed.

davisp (approved):
+1

PR #6, opened 2024-03-18
Finish the Attribute API wrapping

davisp:
The major change here is to reimplement the DomainType as a new `CAPIConverter` trait along with a generic method on `tiledb_sys::Datatype`.


src/convert.rs
```diff
@@ -0,0 +1,42 @@
+pub trait CAPIConverter {
```
rroelke, line 1: I'm still not sure this trait gets it right.  I think we need to play with strings to find a trait that does.  Which is to say, this isn't really any different than what was here before, and I believe it will have to evolve again in the near future.

rroelke: I think I said this in slack already but we will need to distinguish between types which are trivially convertible (as each of this examples is) and types which are not (likely string).  For the trivially convertible cases, we don't want any functions at all, just a marker - and then for the non-trivial cases we will end up with implementations that are more complicated than this.

rroelke, line 1: If I may quibble with the trait name, the types implementing the trait are convertees, not converters :)

tiledb-sys/src/datatype.rs
```diff
@@ -17,6 +18,7 @@ extern "C" {
     pub fn tiledb_datatype_size(type_: u32) -> u64;
 }
 
+#[allow(non_snake_case)]
```
rroelke, line 21: It's a bigger change but I kind of think this should change to be more similar to the smaller enums I've worked through - i.e. the `sys` library just has the constants and then we put the enum in the main crate.  We shouldn't need this.

tiledb-sys/src/datatype.rs
```diff
@@ -207,6 +209,96 @@ impl Datatype {
             _ => None,
         }
     }
+
+    pub fn is_valid<T: 'static>(&self, _: &T) -> bool {
+        if TypeId::of::<T>() == TypeId::of::<f32>() {
```
rroelke, line 214: I think a tuple-type `match (self, TypeId::of::<T>())` could probably neaten this a bit.

tiledb-sys/src/datatype.rs
```diff
@@ -207,6 +209,96 @@ impl Datatype {
             _ => None,
         }
     }
+
+    pub fn is_valid<T: 'static>(&self, _: &T) -> bool {
```
rroelke, line 213: "valid" meaning "`T` is a physical type which implements this logical type"?
I'd probably name it `is_physical_type` or at the least give put the "physical type" terminology in a doc comment

tiledb-sys/src/datatype.rs
```diff
@@ -224,10 +316,80 @@ impl Display for Datatype {
...
-            println!("{}", i);
+            println!("I: {}", i);
+            if i <= 43 {
+                let dt = Datatype::from_u32(i as u32)
+                    .expect("Error converting value to Datatype");
+                assert_ne!(
+                    format!("{}", dt),
+                    "<UNKNOWN DATA TYPE>".to_string()
+                );
+                assert!(check_valid(&dt));
```
rroelke, line 332: I like the assertion that only one physical type is valid for each logical type but the way this is written leaves a lot of room for failure if the wrong one is chosen.  I think it's probably worth the map from logical to physical type (which we probably want somewhere anyway?) and then having the `check_valid` instead return a vector representing which types were valid.

src/array/attribute.rs
```diff
@@ -383,4 +445,33 @@ mod test {
...
+
+        assert!(attr.set_fill_value(&ctx, 5_i32).is_err());
+        assert!(attr.set_fill_value(&ctx, 5_u32).is_err());
+        assert!(attr.set_fill_value(&ctx, 1.0_f64).is_err());
+
+        Ok(())
+    }
+
+    #[test]
+    fn attribute_test_set_fill_value() -> TileDBResult<()> {
```
rroelke, line 463: We will definitely want string tests soon but I think we can skip for now - I left a related comment in `convert.rs`.

rroelke (approved):
Most of my comments are not imminently actionable so I'm choosing to approve rather than request changes. But I think we definitely want the test improvement sooner rather than later.

tiledb-sys/src/datatype.rs
davisp: Not sure what you mean here. Self would be the enum value in this case which is just a variant of an enum rather than the data being converted.

tiledb-sys/src/datatype.rs
davisp: This one I'm not even sure why it started complaining when it was there before the PR. The linter disable was purely to tell clippy to quiet up so I didn't have to include tugging on this loose end.

tiledb-sys/src/datatype.rs
davisp: I had some vague ideas on how this would work with slices that we'll need for strings and multi-value cells. I'll spend a bit this morning setting up an example to see if I can't decide whether we'll be able to use this same pattern for everything or if we'll have to get more complicated.

Which is to say, I'm not 100% that this will only work for physical types. Though I could see renaming to `is_compatible_type` or something.

tiledb-sys/src/datatype.rs
davisp: For sure. The only guarantee that we've picked the right type is to have the second map but I didn't have the brain power to figure out how best to do that in Rust. I'm more than open to adding it if you've got any idea on how to write it.

src/array/attribute.rs
```diff
@@ -234,6 +235,67 @@ impl Attribute {
...
+
+    pub fn get_fill_value<Conv: CAPIConverter>(
+        &self,
+        ctx: &Context,
+    ) -> TileDBResult<Conv> {
+        let mut c_ptr: *const std::ffi::c_void = out_ptr!();
+        let mut c_size: u64 = 0;
+
+        let res = unsafe {
+            ffi::tiledb_attribute_get_fill_value(
```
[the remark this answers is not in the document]
davisp: I don't want to go too far Rust-ifying our interface until more of the library is wrapped so that we have a better feel for common patterns, but I wonder if we couldn't do something along the lines of having a specific type for the value would be the way to go. Something along the lines of:

```
enum FillValue {
    NonNullable(T),
    Nullable(T, B)
}

fun set_fill_value(value: FillValue);
fun get_fill_value() -> FillValule;
```

src/convert.rs
davisp: The conversion stuff isn't any different for sure. The issue was I focusing on was removing the Datatype dependence of the conversion. With get_fill_value I ran smack into the issue when I realized that DomainType was wrong because u64 would only have one associated Datatype via DomainType::DATATTYPE which was wrong.

davisp: The two hardest things in computer science: Naming things, cache invalidation, and off by one errors.

Maybe `CAPIConvertible`?

tiledb-sys/src/datatype.rs
davisp: Honestly, I've got no idea why clippy suddenly started yelling about this. I just did this to silence the warning so I didn't have to tug on the string to fix it in the context of this PR.

src/convert.rs
rroelke: Ah that makes sense.  Yeah multiple datatypes backed by u64.  Got it.

rroelke: Works for me

tiledb-sys/src/datatype.rs
rroelke: Does clippy run on all directories?  If not maybe that's why it didn't catch this.

tiledb-sys/src/datatype.rs
rroelke: I'm happy to check out this branch and give it a whirl, or do so after you've merged

tiledb-sys/src/datatype.rs
davisp: Experimental results show that we should be able to apply this same technique to non-physical types as well. According to the TypeId docs, it currently only works for types that are `'static` which I don't even know how it a type wouldn't be static yet. I'll go with `is_compatible_type` for now unless you throw something at me before I finish updating and merging.

PR #8, opened 2024-03-19
Change to virtual workspace with tiledb-api crate, and add tiledb-arrow crate

rroelke:
One task we must complete for the DataFusion integration is convert from a TileDB array schema to an Arrow schema.  Let's put our arrow integrations in a separate crate.  This pull request turns our repo into a "virtual workspace" with multiple sub-crates.  Along the way we move the top-level API source into a new `tiledb-api` sub-crate and add an arrow sub-crate with trivial functionality (to be fleshed out later).


tiledb-arrow/src/datatype.rs
```diff
@@ -0,0 +1,168 @@
...
+        _ => None, // TODO
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    fn datatype_strategy() -> impl Strategy<Value = tiledb::Datatype> {
```
rroelke, line 104: First use of property-based testing!

davisp:
Two main things I'd change: From my limited experience, most multi-crate projects seem to nest crates under a top level `crates/` directory. Second, I'm not a fan of `tiledb-api` for a crate name. I'd much prefer `use tiledb;` vs `use tiledb_api;` since the `_api` seems rather redundant. Though, if we can call the crate `tiledb` while the directory is `tiledb-api` I would be fine with that as well.

Also, for project structure stuff, there's the section in the book that covers simple cases, but for the more complex multi-crate stuff I tend to use ripgrep as a reference: https://github.com/BurntSushi/ripgrep

rroelke:
> Two main things I'd change: From my limited experience, most multi-crate projects seem to nest crates under a top level `crates/` directory. Second, I'm not a fan of `tiledb-api` for a crate name. I'd much prefer `use tiledb;` vs `use tiledb_api;` since the `_api` seems rather redundant. Though, if we can call the crate `tiledb` while the directory is `tiledb-api` I would be fine with that as well.
> 
> Also, for project structure stuff, there's the section in the book that covers simple cases, but for the more complex multi-crate stuff I tend to use ripgrep as a reference: https://github.com/BurntSushi/ripgrep

I intended to follow the example set by datafusion, but it looks like I missed a step.
Datafusion has all the crates nested in `$ROOT/datafusion` rather than a `crates` directory.

Datafusion exports its crate `datafusion/core` as `datafusion` so we should be able to follow that example for tiledb as well.

davisp (approved):
+1

PR #9, opened 2024-03-19
Initial CI Workflows

davisp:
This adds the initial basic CI runs for PRs and nightlies.

For the moment all versions are currently hardcoded URLs pointing at 2.21.0-rc1. The next RC or final release will include a new `releases.csv` which we can parse so that we can start to parameterize on versions more easily. We'll update these workflows when that's available.


rroelke:
You mentioned occasionally deleting the `Cargo.lock`, is that going to be part of nightly or something else?

davisp:
@rroelke Instead of deleting it which seemed like it could get messy, I ended up just forcing the update via `cargo update --recursive` on the nightly CI

tiledb/api/src/context.rs
```diff
@@ -192,7 +192,13 @@ mod tests {
         // enabled. These tests failing most likely means a non "standard"
         // build of libtiledb.{so,dylib,dll}
         assert!(ctx.is_supported_fs(ffi::Filesystem::MEMFS));
-        assert!(!ctx.is_supported_fs(ffi::Filesystem::HDFS));
+
+        // On GitHub Actions, we use the release tarball which enables all
+        // backends. Thus we skip this test when running in CI.
```
rroelke, line 197: FWIW all of these tests could fail environmentally (in principal).  Poor HDFS getting singled out.

rroelke (approved):

PR #10, opened 2024-03-19
Add separate `test` crate for test utilities

rroelke:
We'll have at least two sub-crates where it will be useful to share common test code (for example, to generate an arbitrary data type for property-based testing).

Here we add yet another sub-crate to encapsulate the testing functionality - `test`.  This crate is meant to be used as a `dev-dependency` in crates which want to access the testing functionality.


davisp (approved):
Hell yeah! More of this please!

PR #11, opened 2024-03-19
Remove test crate circular dependency and add arbitrary Attribute pbt strategy

rroelke:
The addition of the Attribute pbt strategy revealed that the circular dependency did not actually work.

### DIFF
--- a/tiledb/api/src/array/attribute.rs
+++ b/tiledb/api/src/array/attribute.rs
@@ -1,6 +1,7 @@
 extern crate tiledb_sys as ffi;
 
 use std::convert::From;
+use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::ops::Deref;
 
 pub use tiledb_sys::Datatype;
@@ -40,6 +41,10 @@ pub struct Attribute<'ctx> {
 impl<'ctx> Attribute<'ctx> {
     pub(crate) fn as_mut_ptr(&self) -> *mut ffi::tiledb_attribute_t {
         *self.raw
+    }
+
+    pub(crate) fn new(context: &'ctx Context, raw: RawAttribute) -> Self {
+        Attribute { context, raw }
     }
 
     pub fn name(&self) -> TileDBResult<String> {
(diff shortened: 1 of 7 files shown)


tiledb/api/src/array/attribute.rs
```diff
@@ -42,6 +43,10 @@ impl<'ctx> Attribute<'ctx> {
         *self.raw
     }
 
+    pub(crate) fn new(context: &'ctx Context, raw: RawAttribute) -> Self {
```
rroelke, line 46: This change was incidental to match style of previous changes.

tiledb/api/src/array/attribute.rs
```diff
@@ -201,12 +206,23 @@ impl<'ctx> Attribute<'ctx> {
     }
 }
 
-impl<'ctx> From<(&'ctx Context, RawAttribute)> for Attribute<'ctx> {
-    fn from(value: (&'ctx Context, RawAttribute)) -> Self {
-        Attribute {
-            context: value.0,
-            raw: value.1,
-        }
+impl<'ctx> Debug for Attribute<'ctx> {
```
rroelke, line 209: PBT strategy requires Debug.

I chose to emit some fields as JSON since I firmly believe that this sort of thing should be easily processed by a script.

tiledb/arrow/src/datatype.rs
```diff
@@ -92,6 +92,28 @@ pub fn tiledb_type_physical(
         arrow_schema::DataType::UInt64 => Some(tiledb::Datatype::UInt64),
         arrow_schema::DataType::Float32 => Some(tiledb::Datatype::Float32),
         arrow_schema::DataType::Float64 => Some(tiledb::Datatype::Float64),
+        arrow_schema::DataType::Timestamp(
```
rroelke, line 95: Not really related to the PR description but it ended up in there on the branch before I had to deal with the circular dependency.

These cases were found by PBT, as it happens.

tiledb/api/src/array/attribute.rs
davisp: Definitely agree on JSON. Hand generating for the moment seems fine. I'm not sure how heavy of a dependency it is, but I'm guessing we'll probably get some side eye from not using serde_json eventually.

davisp (approved):
+1

PR #15, opened 2024-03-20
Add proptest strategies for generating Dimension, a bunch of Debug impls, and even some arrow stuff

rroelke:
This is what fell out of me working on the Arrow schema conversion functions.  I wanted to do PBT for them, but that means we have to have strategies to generate Schema, which requires Domain, which requires Dimension, and maybe more!

So this pull request ends up being mostly test stuff with a little bit of arrow.

A few notable things:
- the macro `fn_typed` which emulates the same function we have in C++
- the `LifetimeStrategy` struct which works around using `fn_typed` in a strategy and also having a non-static lifetime due to needing the `&'ctx tiledb::context::Context`
- the `Debug` impls I added use `serde_json` to format their results


tiledb/api/src/datatype.rs
```diff
@@ -0,0 +1,161 @@
+#[macro_export]
+macro_rules! fn_typed {
+    ($func:ident, $datatype:expr$(, $arg:expr)* => $then:expr) => {{
```
rroelke, line 3: Trying to mimic
```
template <class Fn, class... Args>
inline auto apply_with_type(Fn&& f, Datatype type, Args&&... args)
```
without variadic generics.

This first rule supports `fn_typed!(generic_function, my_datatype, arg1, arg2, ..., argN => /* do stuff */)`.  The only arguable wart is the `let generic_function = generic_function(arg1, ..., argN);` name shadowing but I don't really mind it much.

tiledb/api/src/datatype.rs
```diff
@@ -0,0 +1,161 @@
...
+        type Datatype = $crate::Datatype;
+        match $datatype {
+            Datatype::Int8 => {
+                let $func = $func::<i8>();
```
rroelke, line 7: Args missing, oops

tiledb/api/src/datatype.rs
```diff
@@ -0,0 +1,161 @@
...
+            Datatype::Boolean => unimplemented!(),
+            Datatype::GeometryWkb => unimplemented!(),
+            Datatype::GeometryWkt => unimplemented!(),
+        }
+    }};
+    ($obj:ident.$func:ident, $datatype:expr$(, $arg:expr)* => $then:expr) => {{
+        type Datatype = $crate::Datatype;
+        match $datatype {
+            Datatype::Int8 => {
+                let $func = $obj.$func::<i8>();
```
rroelke, line 86: Args missing, oops

tiledb/test/src/dimension.rs
```diff
@@ -0,0 +1,101 @@
...
+use proptest::prelude::*;
+use std::fmt::Debug;
+use tiledb::array::{Dimension, DimensionBuilder};
+use tiledb::context::Context;
+use tiledb::fn_typed;
+use tiledb::Result as TileDBResult;
+
+use crate::strategy::LifetimeBoundStrategy;
+
+fn arbitrary_range_and_extent<T>() -> impl Strategy<Value = ([T; 2], T)>
```
rroelke, line 11: There are still some narrow cases where this strategy results in a panic... hopefully I can figure it out :/

davisp (approved):
Awesome work. Most of my comments were just style related. Overall its great though. The only thing I'd say to change is copying your PR comment next to that big macro definition since it really does help clarify what's going on there.

tiledb/arrow/src/attribute.rs
```diff
@@ -0,0 +1,65 @@
...
+    use proptest::prelude::*;
+
+    #[test]
+    fn test_invertibility() -> TileDBResult<()> {
+        let c: TileDBContext = TileDBContext::new()?;
+
+        proptest!(|(attr in tiledb_test::attribute::arbitrary(&c))| {
+            if let Some(arrow_field) = arrow_field(&attr).expect("Error reading tiledb attribute") {
+                assert_eq!(attr.name()?, *arrow_field.name());
+                assert!(crate::datatype::is_same_physical_type(&attr.datatype()?, arrow_field.data_type()));
```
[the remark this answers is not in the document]
rroelke: It definitely seems like rustfmt doesn't know what to do with macros

tiledb/api/src/array/dimension.rs
```diff
@@ -100,6 +103,18 @@ impl<'ctx> Dimension<'ctx> {
     }
 }
 
+impl<'ctx> Debug for Dimension<'ctx> {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        let json = json!({
+            "datatype": format!("{}", self.datatype()),
+            "domain": fn_typed!(self.domain, self.datatype() => match domain { Ok(x) => format!("{:?}", x), Err(e) => format!("<{}>", e) }),
```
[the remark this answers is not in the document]
rroelke: This is what `rustfmt` gives but I edited this a bit to make it a little cleaner.

tiledb/test/src/dimension.rs
rroelke: I think I got it.  For ints at least.  I'm definitely worried about floats.

PR #18, opened 2024-03-21
Implement strategy for arbitrary Schema generation, for dense or sparse arrays

rroelke:
There are some restrictions on dimension types for dense arrays, for which adding some `Datatype` query functions was very helpful.

### DIFF
--- a/tiledb/test/src/schema.rs
+++ b/tiledb/test/src/schema.rs
@@ -1,0 +1,49 @@
+use proptest::prelude::*;
+use tiledb::array::{ArrayType, Schema, SchemaBuilder};
+use tiledb::context::Context;
+use tiledb::Result as TileDBResult;
+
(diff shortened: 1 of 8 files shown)


tiledb/test/src/schema.rs
```diff
@@ -0,0 +1,49 @@
+use proptest::prelude::*;
+use tiledb::array::{ArrayType, Schema, SchemaBuilder};
+use tiledb::context::Context;
+use tiledb::Result as TileDBResult;
+
+pub fn arbitrary_array_type() -> impl Strategy<Value = ArrayType> {
+    // TODO: implement Domain scenario for Dense arrays
```
rroelke, line 7: Oops, can delete comment.

davisp (approved):
+1

PR #19, opened 2024-03-21
Rewrap Filter and FilterList

davisp:
This updates the first APIs I wrapped to use the both the builder pattern as well as tracking their own context references.

### DIFF
--- a/tiledb/api/src/filter.rs
+++ b/tiledb/api/src/filter.rs
@@ -1,19 +1,27 @@
-extern crate tiledb_sys as ffi;
-
 use std::ops::Deref;
-
-pub use tiledb_sys::Datatype;
-pub use tiledb_sys::FilterOption;
-pub use tiledb_sys::FilterType;
-pub use tiledb_sys::WebPFilterInputFormat;
 
 use crate::context::Context;
 use crate::error::Error;
 use crate::Result as TileDBResult;
 
+pub enum CompressionType {
+    Bzip2,
+    Delta,
+    Dictionary,
+    DoubleDelta,
+    Gzip,
+    Lz4,
+    Rle,
+    Zstd,
+}
+
+pub enum ChecksumType {
+    Md5,
+    Sha256,
+}
+
 pub(crate) enum RawFilter {
     Owned(*mut ffi::tiledb_filter_t),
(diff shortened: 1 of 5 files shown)


tiledb/api/src/filter.rs
```diff
@@ -1,345 +1,216 @@
...
-            ctx,
-            FilterOption::WEBP_INPUT_FORMAT,
-            &c_format as *const std::ffi::c_uchar as *const std::ffi::c_void,
+            ffi::FilterOption::POSITIVE_DELTA_MAX_WINDOW,
+            &mut c_width as *mut std::ffi::c_uint as *mut std::ffi::c_void,
         )
+        .map(|_| c_width as u32)
     }
 
     pub fn get_webp_input_format(
```
rroelke, line 167: This being in the `Filter` impl for all of the filter types does point to this being one of multiple steps to clean this up.

tiledb/api/src/filter.rs
```diff
@@ -1,345 +1,216 @@
...
         &self,
-        ctx: &Context,
-    ) -> TileDBResult<WebPFilterInputFormat> {
+    ) -> TileDBResult<ffi::WebPFilterInputFormat> {
```
rroelke, line 169: Hm, we probably want to avoid having `ffi::` in the public APIs.

rroelke (approved):
I left a few notes but those are things we can improve later.  This looks like a great step

tiledb/api/src/filter.rs
davisp: Absolutely. I'm sure there's a better way to handle this, but it felt like reaching for `Boxy<dyn Something>` which is beyond where I want to go at this stage.

The only thing that makes me ok with the status quo is knowing that these particular API's will likely only be used by use for implementing `Debug` type traits when folks are just looking for a description of the state.

tiledb/api/src/filter.rs
davisp: I definitely wasn't fond of that either, but kinda left it as a "We'll clean up the api vs sys symbols at some point". For the interim, I'll go add a public use to just import `WebPFilterInputFormat` into the `api/filter` namespace to remove the `ffi` tokens.

PR #20, opened 2024-03-21
Rename as_mut_ptr() to capi() for consistency

davisp:
First, all instances of as_mut_ptr() have been replaced with the now standard capi(). Both of these do exactly the same thing, this just unifies the naming convention.

Secondly, I've moved almost everything over to using the RawThing/Thing pattern where RawThing is just a deref-able enum containing the ffi pointer. This removes the need for all of the instances of as_mut_ptr_ptr and Default implementations that were sketchy at best.

Filters haven't been updated yet, but they will be as part of PR #19.


rroelke (approved):
Thanks!

PR #21, opened 2024-03-22
impl PartialEq for Attribute

rroelke:
To test Arrow schema to/from TileDB schema conversion, we will want to test invertability.  That will require some means of checking that the Attribute we get out is the same as the Attribute we put in.  `PartialEq` is a natural choice for this, so this pull request does most of the work to implement it for Attribute.

What's missing is the filter comparison, and that will probably be done as a follow-up after #19 is merged.

### DIFF
--- a/tiledb/api/src/array/attribute.rs
+++ b/tiledb/api/src/array/attribute.rs
@@ -8,9 +8,10 @@ use serde_json::json;
 pub use tiledb_sys::Datatype;
 
 use crate::context::Context;
-use crate::convert::CAPIConverter;
+use crate::convert::{BitsEq, CAPIConverter};
 use crate::error::Error;
 use crate::filter_list::{FilterList, RawFilterList};
+use crate::fn_typed;
 use crate::Result as TileDBResult;
 
 pub(crate) enum RawAttribute {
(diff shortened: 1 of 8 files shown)


tiledb/api/src/convert.rs
```diff
@@ -29,3 +29,52 @@ impl<T: CAPISameRepr> CAPIConverter for T {
         *value
     }
 }
+
+/// Trait for comparisons based on value bits.
+/// This exists to work around float NaN which is not equal to itself,
```
rroelke, line 34: I would love, love, love if we did not need this.

Fun fact:
`impl<T> BitsEq for T where T: Eq` is forbidden in concert with `impl BitsEq for f32` because the compiler says that `std` may impl Eq for f32 someday.  No it won't.

tiledb/api/src/error.rs
```diff
@@ -98,4 +100,13 @@ impl Drop for Error {
     }
 }
 
+impl Serialize for Error {
```
rroelke, line 103: Arguably this is questonable, but it makes the `json!` in the `Debug` impls way more concise.  Although maybe a generic function would work instead actuallly...

tiledb/test/src/attribute.rs
```diff
@@ -30,4 +32,13 @@ mod tests {
 
         proptest!(|(_ in arbitrary(&ctx))| {});
     }
+
+    #[test]
+    fn attribute_eq_reflexivity() {
+        let ctx = Context::new().expect("Error creating context");
+
+        proptest!(|(attr in arbitrary(&ctx))| {
+            assert_eq!(attr, attr);
```
rroelke, line 41: This is what caught the `NaN` thing.

tiledb/api/src/array/attribute.rs
```diff
@@ -210,23 +211,110 @@ impl<'ctx> Attribute<'ctx> {
...
         write!(f, "{}", json)
     }
 }
 
+impl<'c1, 'c2> PartialEq<Attribute<'c2>> for Attribute<'c1> {
+    fn eq(&self, other: &Attribute<'c2>) -> bool {
+        let names_match = match (self.name(), other.name()) {
+            (Ok(mine), Ok(theirs)) => mine == theirs,
+            _ => false,
+        };
```
davisp, line 256: Oooh, this is awesome!

tiledb/api/src/error.rs
davisp: I'm only confused why you think its questionable. I mean, we could get fancier with making it an object with richer information, but as far ask I can tell there's no worry about somehow causing a panic.

tiledb/api/src/convert.rs
davisp: i'd actually add this as a comment before merging. Without having had the context of knowing there was a NaN issue to begin with, this particular change without context might be a bit odd. And in the crazy case they ever do implement `Eq` for `f32` we'll have the confidence to remove it.

davisp (approved):
+1 As per usual, my only note is to add your PR comment as an actual comment so that code doesn't lose context. The choice makes perfect sense with the NaN issue, but I could see that connection being lost over time.

tiledb/api/src/error.rs
rroelke: This is how I want to format the error messages when they would show up in `Debug`, but that doesn't mean it's how we should serialize an error message.  Whereas this conflates the two.

PR #23, opened 2024-03-22
Scaffolding to support generating arbitrary filter pipelines

rroelke:
The missing piece from `Attribute::eq` was the filter pipeline.  To get this in there with the testing infrastructure we have and/or will need, we need to do `Debug`, `Eq`, `Serialize` and `Deserialize` for `FilterList` and `Filter`.

That turns out to require a bunch of things:
- to keep the `sys` crate with minimal dependencies, I moved `Datatype` to the `api` crate where it can `#[derive(Deserialize, Serialize)]`.  This also is a big step towards keeping `ffi::` out of the public APIs.
- not strictly a requirement, but the `enum FilterData` is really useful because it can `#[derive(PartialEq)]` which helps prevent us developers from making mistakes.  Now those mistakes can live in `Filter::create` and/or `Filter::filter_data` instead.
- `FilterData::transform_datatype` is an internal C API, but we need to use it for property-based testing to make sure that we construct valid filter pipelines.  The output of each filter in the pipeline needs to be a valid input to the next one.

### DIFF
--- a/tiledb/api/src/array/attribute.rs
+++ b/tiledb/api/src/array/attribute.rs
@@ -5,14 +5,13 @@ use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::ops::Deref;
 
 use serde_json::json;
-pub use tiledb_sys::Datatype;
 
 use crate::context::Context;
 use crate::convert::{BitsEq, CAPIConverter};
 use crate::error::Error;
 use crate::filter_list::{FilterList, RawFilterList};
 use crate::fn_typed;
-use crate::Result as TileDBResult;
+use crate::{Datatype, Result as TileDBResult};
 
 pub(crate) enum RawAttribute {
     Owned(*mut ffi::tiledb_attribute_t),
@@ -70,11 +69,7 @@ impl<'ctx> Attribute<'ctx> {
             ffi::tiledb_attribute_get_type(c_context, *self.raw, &mut c_dtype)
         };
         if res == ffi::TILEDB_OK {
-            if let Some(dtype) = Datatype::from_u32(c_dtype) {
-                Ok(dtype)
-            } else {
-                Err(Error::from("Invalid Datatype value returned by TileDB"))
-            }
+            Datatype::try_from(c_dtype)
         } else {
             Err(self.context.expect_last_error())
         }
@@ -238,13 +233,10 @@ impl<'ctx> Debug for Attribute<'ctx> {
                 } else {
                     None
                 },
-            /*
-            TODO
             "filters": match self.filter_list() {
                 Ok(fl) => format!("{:?}", fl),
                 Err(e) => format!("<error reading filters: {}>", e)
             },
-            */
             "raw": format!("{:p}", *self.raw)
         });
         write!(f, "{}", json)
@@ -274,15 +266,13 @@ impl<'c1, 'c2> PartialEq<Attribute<'c2>> for Attribute<'c1> {
             return false;
         }
 
-        /*
         let filter_match = match (self.filter_list(), other.filter_list()) {
-            (Ok(mine), Ok(theirs)) => unimplemented!(),
+            (Ok(mine), Ok(theirs)) => mine == theirs,
             _ => false,
         };
         if !filter_match {
             return false;
         }
-        */
 
         let cell_val_match = match (self.cell_val_num(), other.cell_val_num()) {
             (Ok(mine), Ok(theirs)) => mine == theirs,
@@ -545,12 +535,17 @@ mod test {
         {
             let flist2 = FilterListBuilder::new(&ctx)
                 .expect("Error creating filter list builder.")
-                .add_filter(NoopFilterBuilder::new(&ctx)?.build())?
-                .add_filter(BitWidthReductionFilterBuilder::new(&ctx)?.build())?
-                .add_filter(
-                    CompressionFilterBuilder::new(&ctx, CompressionType::Zstd)?
-                        .build(),
-                )?
+                .add_filter(Filter::create(&ctx, FilterData::None)?)?
+                .add_filter(Filter::create(
+                    &ctx,
+                    FilterData::BitWidthReduction { max_window: None },
+                )?)?
+                .add_filter(Filter::create(
+                    &ctx,
+                    FilterData::Compression(CompressionData::new(
+                        CompressionType::Zstd,
+                    )),
+                )?)?
                 .build();
 
             let attr = Builder::new(&ctx, "foo", Datatype::UInt8)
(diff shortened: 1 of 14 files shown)


rroelke:
This doesn't quite work yet, the `Attribute` proptest is failing to construct because the pipeline viability code I wrote is not quite right.

tiledb/api/src/filter.rs
```diff
@@ -54,26 +274,139 @@ impl<'ctx> Filter<'ctx> {
...
+                    Self::set_option(
+                        context,
+                        *raw,
+                        ffi::FilterOption::COMPRESSION_LEVEL,
+                        c_level,
+                    )?;
+                }
+                if let Some(reinterpret_datatype) = reinterpret_datatype {
+                    let c_datatype =
+                        reinterpret_datatype.capi_enum() as std::ffi::c_uchar;
```
davisp, line 326: is `std::ffi::c_uchar` correct here? I would have expected `u32` or `std::ffi::c_uint`.

tiledb/api/src/filter.rs
```diff
@@ -82,445 +415,185 @@ impl<'ctx> Filter<'ctx> {
...
-        match res {
-            Ok(()) => {
-                match ffi::WebPFilterInputFormat::from_u32(c_fmt as u32) {
-                    Some(fmt) => Ok(fmt),
-                    None => Err(Error::from(
-                        "Invalid WebP input filter format returned from core.",
-                    )),
-                }
+        let get_compression_data =
+            |compression_type| -> TileDBResult<FilterData> {
```
davisp, line 423: Style nit: I imagine this would be a whole lot easier to read if we renamed the argument to `kind` and then assigned the level and reinterpret type to temporaries so that the end is just:

```rs
Ok(FilterData::Compression(CompressionData { kind, level, reinterpret_datatype }))
```

tiledb/api/src/filter.rs
```diff
@@ -15,11 +19,227 @@ pub enum CompressionType {
...
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum FilterData {
+    None,
+    BitShuffle,
+    ByteShuffle,
+    BitWidthReduction {
+        max_window: Option<u32>,
```
davisp, line 90: Why is just this setting optional?

tiledb/api/src/filter.rs
```diff
@@ -82,445 +415,185 @@ impl<'ctx> Filter<'ctx> {
...
+                            dtype as ffi::tiledb_datatype_t,
+                        ) {
+                            Ok(dtype) => dtype,
+                            Err(_) => {
+                                return Err(Error::from(format!(
+                                "Invalid compression reinterpret datatype: {}",
+                                dtype
+                            )))
+                            }
+                        }
```
davisp, line 443: Maybe use `map_err` here? Something along the lines of:

```rs
Datatype::try_from(dtype as ffi::tiledb_datatype_t).map_err(|e| {
    Error::from(format!("Invalid compression reinterpret datatype '{}': {}", dtype, e))
})?;
```

tiledb/api/src/filter.rs
```diff
@@ -82,445 +415,185 @@ impl<'ctx> Filter<'ctx> {
...
-impl<'ctx> ChecksumFilterBuilder<'ctx> {
-    pub fn new(
-        context: &'ctx Context,
-        checksum_type: ChecksumType,
-    ) -> TileDBResult<Self> {
-        let ftype = match checksum_type {
-            ChecksumType::Md5 => ffi::FilterType::ChecksumMD5,
-            ChecksumType::Sha256 => ffi::FilterType::ChecksumSHA256,
+        /*
+        let types_match = match (self.get_type(), other.get_type()) {
```
davisp, line 585: What's this bit? Looks like it was some left over WIP that just needs removing?

tiledb/api/src/filter_list.rs
```diff
@@ -67,6 +68,12 @@ impl<'ctx> FilterList<'ctx> {
         }
     }
 
+    pub fn to_vec(&self) -> TileDBResult<Vec<Filter<'ctx>>> {
+        (0..self.get_num_filters()?)
+            .map(|f| self.get_filter(f))
```
davisp, line 73: Huh. TIL that collect() returns the first error or the whole list. That's awesome!

davisp (commented):
Whoa. Yeah, that' turned out way better than I was expecting when we discussed possible approaches on the Thursday call. It really shines when looking at the proptest definitions as well.

Most things I had were either minor style nits or general questions that don't need answering for this PR. The only real thing I saw was that `c_uchar` vs `c_uint` question.

Otherwise, +1 once CI is green.

tiledb/api/src/filter.rs
davisp: That's a bit ambiguous in hindsight. I meant to ask, of all the options for all the filter types, why is just this max_window optional?

tiledb/api/src/filter.rs
rroelke: Surprisingly yes - it expects just one byte inside of libtiledb

tiledb/api/src/array/attribute.rs
```diff
@@ -458,7 +448,7 @@ impl<'ctx> Builder<'ctx> {
         }
     }
 
-    pub fn filter_list(self, filter_list: &FilterList) -> TileDBResult<Self> {
+    pub fn filter_list(self, filter_list: FilterList) -> TileDBResult<Self> {
```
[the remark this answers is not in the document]
rroelke: I'm happy to revert this. I'm not really sure why I changed it to be honest.

tiledb/api/src/filter.rs
rroelke: The answer is mostly laziness, this is the only one that came up in the tests where I had to actually write down the value.  And of course I didn't know it, so I decided to go on to option type.

Really all of these or none of these should be option.

tiledb/api/src/array/attribute.rs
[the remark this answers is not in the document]
davisp: I vote revert now, though if you want to wait to see if it becomes an issue as we start writing more examples and tests I'm fine with that. I just really expect that wanting to pass the same filter pipeline to multiple attributes is a thing we'll want to support for ergonomic reasons.

tiledb/api/src/filter.rs
davisp: Well that's awkward. In a note to future selves, I tracked this down to the fact that most of our uses of Datatype refelect the public C API `typedef enum { ... } tiledb_datatype_t` definition which is apparently backed by a `u32` according to bindgen. However, the `Filter::set_option` pipes the `void*` all the way to `CompressionFilter::set_option_impl` which is casting it to a `tiledb::sm::Datatype` which is `enum class Datatype : uint8_t { ... }` definition which means it really does want a single byte ini this case.

PR #24, opened 2024-03-22
Initial wrappers for Enumerations

davisp:


tiledb/api/src/array/enumeration.rs
```diff
@@ -0,0 +1,502 @@
...
+                    as *mut *const std::ffi::c_void,
+                &mut size,
+            )
+        };
+
+        // Rust docs say that we're not allowed to pass a nullptr to the
+        // std::slice::from_raw_parts because that breaks how Option<&[T]>::None
+        // is represented.
+        if ptr.is_null() {
+            ptr = 0x1 as *const std::ffi::c_uchar;
```
rroelke, line 131: Doesn't this all but guarantee a segfault?  I guess not because the size is zero?

This line gave me a lot of pause, but it does turn out there is an idiomatic way to do this mentioned on the [function reference page](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html):

"Data must be non-null and aligned even for zero-length slices. One reason for this is that enum layout optimizations may rely on references (including slices of any length) being aligned and non-null to distinguish them from other data. You can obtain a pointer that is usable as data for zero-length slices using [NonNull::dangling()](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.dangling)."

tiledb/api/src/array/enumeration.rs
```diff
@@ -0,0 +1,502 @@
...
+        let offsets_ptr = if column.offsets().is_empty() {
+            std::ptr::null_mut()
+        } else {
+            column.offsets().as_ptr()
+        };
+
+        // An important note here is that the Enumeration allocator copies the
+        // contents of data of offsets rather than assumes ownership. That
+        // means this is safe as those bytes are guaranteed to be alive until
+        // we drop self at the end of this method after returning from
```
rroelke, line 193: Is this method meant to consume `self`?  RIght now it is a reference

tiledb/api/src/array/enumeration.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        let ordered_match = match (self.ordered(), other.ordered()) {
+            (Ok(mine), Ok(theirs)) => mine == theirs,
+            _ => false,
+        };
+        if !ordered_match {
+            return false;
+        }
+
+        // TODO: Match data and offsets
```
rroelke, line 301: Probably can fetch the column and `impl PartialEq` for the column, that would be a good way to do it.

tiledb/api/src/array/enumeration.rs
```diff
@@ -0,0 +1,502 @@
...
+        // as a valid strategy. For our situation, we just use a zero length to
+        // indicate when we should pass nullptr.
+        let offsets_ptr = if column.offsets().is_empty() {
+            std::ptr::null_mut()
+        } else {
+            column.offsets().as_ptr()
+        };
+
+        // An important note here is that the Enumeration allocator copies the
+        // contents of data of offsets rather than assumes ownership. That
```
rroelke, line 362: of data *and* offsets? is that what you mean?

rroelke (changes requested):
I left a bunch of comments, some questions, some comments, some flags for follow-up... the ones I marked with :eyes: are the ones that I think should be improved before merging

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+/// The Column struct is used to represent a column of data in TileDB. A column
+/// contains a required data slice and an optional offsets slice when the
+/// underlying column definition has a variable number of values per cell.
+///
+/// An AsColumn trait is provided for a number of common use cases such that
+/// users should be able to pass common instances directly as arguments that
+/// have the Column type.
+pub enum Column<'data> {
```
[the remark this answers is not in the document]
davisp: > I'm picturing some kind of marker object which uses the `'data` lifetime to make sure that the user data lives longer than the query does. That probably looks even more different between `ReferenceColumn` and `AllocatedColumn`.

My goal with the `'data` lifetime was for the case of when we're just passing a pointer to Rust allocated RAM into TileDB so that the Column lifetime would be guaranteed to be shorter than whatever we might have borrowed from. I've got no idea if I've gotten that correct. So I should probably add a test for it specifically. And as you mentioned, those `compile_fail` tests seem like the best practice for doing that.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,416 @@
...
+        if !datatype.is_compatible_type::<T>() {
+            return Err(Error::from(format!(
+                "Requested datatype is incompatible with {}",
+                datatype.to_string().unwrap_or("<unknown type>".to_owned())
+            )));
+        }
+
+        // For var sized data, use the to_type_vec methods for conversion.
+        if cell_val_size == u32::MAX {
+            return Err(Error::from(
```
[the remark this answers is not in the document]
davisp: I'm not entirely sure which bit you're referring to here. The `cell_var_size == u32::MAX` equating to var sized-ness or the `Err(Error::from(...))` syntax.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        Ok(ret)
+    }
+
+    pub fn to_boxed_u8_vec(
+        &self,
+        datatype: Datatype,
+        cell_val_size: u32,
+    ) -> TileDBResult<Vec<Box<[u8]>>> {
+        // Currently we're only supporting ASCII and UTF-8 strings.
```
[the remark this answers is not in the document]
davisp: It looks like UTF-16 has decent support. It looks like we'd have to write our own UTF-32/UCS-2/UCS-4 converters based on the standard library. And/or just add a dependency if we can find something specific enough to be worth the tradeoff there.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,416 @@
...
+    offsets: &'data [u64],
+}
+
+/// If a column requires allocating a linearized buffer (i.e., when passing a
+/// Vec or slice of AsRef<str> values), this structure is used to hold the
+/// allocated buffers.
+pub struct AllocatedColumn {
+    num_values: u64,
+    value_size: u64,
+    data: Box<[u8]>,
```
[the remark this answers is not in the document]
davisp: Not sure which copy you're considering here?

The main reason I went for `Box<[u8]>` is that I was attempting to find a way to just create a large fixed sized buffer in Rust and that was the primary approach I saw while Googling.

However, in hindsight, there are two other decent reasons for preferring it over Vec. First, Vec can be resized which changes it's pointers and I'm not certain that the borrow checker can handle that or not. And if it can't, that seems like a prime recipe for segfaulting when TileDB attempts to use a pointer from a Vec that got resized and invalidated the pointer given to TileDB.

Secondly, probably much less important, but Vec has the size vs capacity issue where a lot of Vec's can end up wasting a lot of RAM due to their memory growth. Its not an issue most of the time, but when it is an issue, it can be a pain to fix.

tiledb/api/src/array/enumeration.rs
davisp: Yep. I wrote these before Column thinking I was gonna go fill out the `arbitrary_enumeration` implementation first. But then it bugged me I wasn't asserting the major differences so went back to figuring out the Column stuff. I hadn't thought of adding PartialEq on the Column itself. That's a great idea. My current plan is to just make sure that the sizes are equal and then figure Rust's idea of `std::memcmp`.

tiledb/api/src/array/enumeration.rs
davisp: Totes, will fix.

tiledb/api/src/array/enumeration.rs
```diff
@@ -0,0 +1,502 @@
...
+                raw: RawTDBString::Owned(c_str),
+            };
+            tdb_str.to_string()
+        } else {
+            Err(self.context.expect_last_error())
+        }
+    }
+
+    pub fn datatype(&self) -> TileDBResult<Datatype> {
+        let mut dtype: ffi::tiledb_datatype_t = 0;
```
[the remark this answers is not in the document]
davisp: Oh, does that work for the non-pointer types? I've certainly been trying to remember to use that for all the pointers, but for anything concrete like u32/u64 I've just been zero-initializing before use.

tiledb/api/src/array/enumeration.rs
davisp: Oooh, good find on that `NonNull::dangling()` to clean this up. I had seen the note about 0x0 vs 0x1 issue so just implemented it directly figuring that all the alignment stuff would Just Work since it was Rust controlling things. However, using the named stdlib approach is even better than the weird looking pointer cast that as you pointed out looks exactly like prime segfault territory even though Rust actually requires it.

I assume you're on board with just doing:

```rust
if ptr.is_null() {
    ptr = std::ptr::NonNull::dangling();
}
```

tiledb/api/src/array/enumeration.rs
davisp: It is specifically not consuming self because of the underlying implementation. `tiledb::sm::Enumeration::extend` creates a new instance of `Enumeration` that gets a copy of the "parent" enumeration's data with a copy of the extension data appended. I.e., it's a copy underneath rather than a sharing of data.

The reasoning here is that this function was added to aid language bindings do schema evolution so the "parent" instance is still in use in its original form which is what prompts the returning a new instance design.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,416 @@
...
+use crate::Result as TileDBResult;
+
+/// The ReferenceColumn is used when we can safely use the existing Vec's or
+/// slices being used to create a column. This avoids the allocation, copy, and
+/// eventual deallocation when providing column data to TileDB.
+pub struct ReferenceColumn<'data> {
+    num_values: u64,
+    value_size: u64,
+    data: &'data [u8],
+    offsets: &'data [u64],
```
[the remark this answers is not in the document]
rroelke: What I like about option is that anyone reading the code is optional.  It avoids questions like "Isn't this not used for most data types?"  :)

So I definitely see a benefit even if there is another clear way to represent "this thing is not used".

tiledb/api/src/column.rs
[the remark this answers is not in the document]
rroelke: Oh, I guess it is kinda bidirectional.
If you get a column back from an Enumeration then the Enumeration should out-live the column.
If you supply a column to a query (possibly where I imagine this going), then the column handle marker thing I allude to should out-live the query.

I think the use of `'data` will be able to accomplish both of these ideas

tiledb/api/src/column.rs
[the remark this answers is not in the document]
rroelke: Since this is the box o' bytes version though we don't really need to deal with the encoding, it's up to whoever the user is going to be.  The line below this is `matches!(datatype, Datatype::StringAscii | Datatype::StringUtf8)` and that can be more permissive to allow the other types.

tiledb/api/src/array/enumeration.rs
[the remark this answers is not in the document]
rroelke: Yep, it does.  Maybe we should change the name to `out_arg!()`.

tiledb/api/src/array/enumeration.rs
rroelke: Yep that works for me

tiledb/api/src/array/enumeration.rs
rroelke: Oh, that makes sense - I was confused because the line break starts with "we drop self at the end of this method", but you don't.

Does this mean you end up with two enumerations with the same name?  Are they also versioned like the array writes are?

tiledb/api/src/column.rs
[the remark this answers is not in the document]
rroelke: I'm thinking about the `string_to_buffers` function principally, so I'll move the comment over there.

In general, the usual rules of borrow checking apply to `Vec` as to anything else - if something has a reference to the interior data, then nothing is allowed to get a mutable reference to modify it.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,416 @@
...
+    };
+}
+
+derive_strings_as_column!(&str);
+derive_strings_as_column!(&String);
+derive_strings_as_column!(String);
+
+// Internal helper to convert a slice of strings to a single linear data buffer
+// and an offsets buffer.
+fn strings_to_buffers(vals: &[impl AsRef<str>]) -> (Box<[u8]>, Box<[u64]>) {
```
rroelke, line 364: This function copies all of the string data.  The alternative I intended to ask about above would be returning a `Vec<&'data [u8]>`, i.e. for each of the input strings a pointer to the data for that string, such that a copy of all the string contents is not necessary.

tiledb/api/src/column.rs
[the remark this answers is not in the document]
davisp: Fair enough. I'll make the switch.

tiledb/api/src/column.rs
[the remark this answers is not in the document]
davisp: Sounds good!

tiledb/api/src/column.rs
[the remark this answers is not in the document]
davisp: That bit is mostly just convenience coupled with the fact that I wasn't thinking too hard since the `to_string_vec` is Generic I was avoiding dealing with more than one `String::from_whatever` calls. Though I could easily add the StringUtf16 variant using conditional logic.

Although, I think your underlying point is more "We should let folks have access to the raw data" which is technically possible with the `to_boxed_u8_vec`, but I could see there being a reasonable request for having `to_u8_slices` to avoid copies for folks wanting to do other things efficiently.

tiledb/api/src/array/enumeration.rs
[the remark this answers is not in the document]
davisp: Lets follow up with this in a separate PR that just normalizes the use in all our API calls. I'm on board in principle, but I want to spend more time reading up on the "correct" way to handle these that we can then standardize on.

tiledb/api/src/array/enumeration.rs
davisp: Correct, the background here is that once users got access to Enumerations, the immediately starting asking to allow for empty Enumerations that are then evolved to have more values. The general idea was that the actual enumeration values aren't known at Array creation time, and are instead only discovered during the first import.

The way it works is before a write, users can evolve the schema with the new Enumeration values. However, the way schema evolution works is basically that we just write a new one and then all new writes will use that instead of the older versions. So the original Enumeration stays for the fragments written using it, and then the new schema has a new copy with the same name (but on disk its a UUID mapping so they don't clash). There's also logic that ensures that names in a given schema are unique, but not necessarily across multiple schemas.

Also, for extra color, the entire extend API is also just a helper to avoid a foot gun in that the values of the prefix must be equivalent for the schema evolution to succeed. Without that you would end up where completely different values are returned depending on which fragments were opened for reading.

tiledb/api/src/column.rs
davisp: Not sure I follow there. The multiple strings version requires a copy because of TileDB's APIs. The way string data is passed is like such:

```rust
// To pass something like this to TileDB
let values = vec!["foo", "bar", "baz"];

// We have to pass this:
let values = "foobarbaz";
let offsets: [u64] = [0, 3, 6];
```

There's also the obvious missing `AsColumn` implementation for tuples so that folks can do things like:

```
let values = "foobarbaz";
let offsets: [u64] = [0, 3, 6];

builder.thing((values, offsets).as_column())
```

I just haven't needed that one yet. And we'll also want to extend columns to support nullable attributes which could then look something like:

```
let values = "foobarbaz";
let offsets: [u64] = [0, 3, 3, 6];
let validity: [u8] = [1, 1, 0, 1];

builder.do_thing((values, offsets, validity).as_column())
```

And I've also suddenly realizing making offsets be u64 unconditionally means that the validity/offsets buffers won't interfere with each other which is nice.

tiledb/api/src/column.rs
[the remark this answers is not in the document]
rroelke: Yes, that's my underlying point - no need to restrict the raw access.  This function, and a `to_u8_slices` if we add one, should allow any of the `StringWhatever` types

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    pub fn is_allocated(&self) -> bool {
+        match self {
+            Column::Referenced(_) => false,
+            Column::Allocated(_) => true,
+        }
+    }
+
+    pub fn to_json(
```
rroelke, line 282: This is probably better done as a `impl Serialize` right?  Then the JSON serializer will pick it up.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,509 @@
...
+            return false;
+        }
+
+        if self.data() != other.data() {
+            return false;
+        }
+
+        let offsets_match = match (self.offsets(), other.offsets()) {
+            (Some(mine), Some(theirs)) => mine == theirs,
+            _ => false,
```
rroelke, line 323: Too strict for non-string columns.  `(None, None)` should also be a match.  You can probably just do `self.offset() == other.offsets()` to be honest

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,509 @@
...
+/// 3. The third is a pair of Vec's or slices where the first is the contents
+///    of the data buffer and the second is a `Vec<u64>` or `&[u64]` that
+///    contains the offsets for each var sized cell. Generally speaking this
+///    is an uncommon use case when users need a variable sized cell that is
+///    not a string.
+pub trait AsColumn<'data> {
+    fn as_column(&'data self) -> Column<'data>;
+}
+
+pub trait TryAsColumn<'data> {
```
rroelke, line 356: This is a more specialized `TryInto` ([doc](https://doc.rust-lang.org/std/convert/trait.TryInto.html)), probably better just to do `impl<'data, T: AsColumn<'data>> TryInto<Column<'data>> for T`

tiledb/api/src/column.rs
rroelke: > The multiple strings version requires a copy because of TileDB's APIs

This was actually something I wondered about. Maybe it doesn't have to. I'll play with it on my own maybe.

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,509 @@
...
+                        self.as_ptr() as *const u8,
+                        num_bytes,
+                    )
+                };
+
+                Column::from_references(
+                    self.len() as u64,
+                    std::mem::size_of::<$typename>() as u64,
+                    data,
+                    Some(&[0u64; 0]),
```
rroelke, line 402: Should be `None` now

tiledb/api/src/array/enumeration.rs
```diff
@@ -246,21 +247,42 @@ impl<'data> TryAsColumn<'data> for Enumeration<'data> {
...
+
+        let dtype_string =
+            dtype.to_string().unwrap_or("<unknown datatype>".to_owned());
         let cell_val_num = self.cell_val_num().map_err(|_| fmt::Error)?;
         let ordered = self.ordered().map_err(|_| fmt::Error)?;
-
-        // TODO: Add enumeration values display.
+        let col = self.try_as_column().map_err(|_| fmt::Error)?;
+
+        let is_strings =
```
rroelke: Wouldn't `StringUcs4` need the offsets and have to do `to_boxed_u8_vec`?
Are attributes of that type sure to have `cell_val_num == u32::MAX`?

tiledb/api/src/column.rs
```diff
@@ -0,0 +1,509 @@
...
+            )));
+        }
+
+        // For var sized data, use the to_type_vec methods for conversion.
+        if cell_val_size != u32::MAX {
+            return Err(Error::from("Use to_vec<T> for fixed sized data."));
+        }
+
+        if self.offsets().is_none() {
+            return Ok(Vec::new());
```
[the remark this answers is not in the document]
davisp: It would be uncommon to not provide them, but an attribute with fixed length strings is something that is supported. Where "fixed length" means same number of bytes.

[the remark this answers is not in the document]
davisp: In other news, I've been waffling over whether or not a Column should know what datatype and cell value number it's representing. It would certainly help in a number of places, while it also feels like it'd unnecessarily restrict where a Column can be used and how they can be created.

tiledb/api/src/column.rs
davisp: Originally I had this bit of logic in the Enumeration Debug trait and realized that it'd make a lot more sense have a Debug for Column trait. And then I ran smack dab into the issue I noted above about Column's not knowing their datatype or cell value number so I had to make it a callable method to use in the Enumeration Debug trait.

tiledb/api/src/column.rs
davisp: Ooof, fair point on None, None being equal. I ran into clippy yelling at me to just directly compare slices and then validated that rustc emits memcmp in its assembly bits. However, when I tried figuring out if two Options are equal all of the suggestions were to use this pattern when its not certain if one or the other is None. Also, it had the standard "eventually `==` will be usable when random language feature X stabilizes" which was written three years ago and the feature is still not stable.

tiledb/api/src/column.rs
davisp: Doh! Good catch.

tiledb/api/src/array/enumeration/strategy.rs
```diff
@@ -0,0 +1,161 @@
...
+
+fn prop_cell_val_num() -> impl Strategy<Value = Option<u32>> {
+    Just(None)
+}
+
+fn prop_ordered() -> impl Strategy<Value = bool> {
+    any::<bool>()
+}
+
+trait WtfNumCmp: std::marker::Sized {
```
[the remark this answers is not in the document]
rroelke: Is NaN allowed in enumerations? I was also curious why the NotNan trait you showed me was of no use here.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+            dtype.to_string().unwrap_or("<unknown datatype>".to_owned());
+        let cell_val_num = self.cell_val_num().map_err(|_| fmt::Error)?;
+        let ordered = self.ordered().map_err(|_| fmt::Error)?;
+
+        let json = json!({
+            "name": name,
+            "datatype": dtype_string,
+            "cell_val_num": cell_val_num,
+            "ordered": ordered,
+            "values": [], // TODO: Render values
```
rroelke, line 239: Other impls of Debug construct the corresponding `ThingData` and then format that. That would get you the value formatting here.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        });
+        write!(f, "{}", json)
+    }
+}
+
+impl<'c1, 'c2> PartialEq<Enumeration<'c2>> for Enumeration<'c1> {
+    fn eq(&self, other: &Enumeration<'c2>) -> bool {
+        let names_match = match (self.name(), other.name()) {
```
rroelke, line 247: I added a macro `eq_helper!` which can condense a lot of the boilerplate here

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+            return false;
+        }
+
+        true
+    }
+}
+
+pub struct Builder<'ctx> {
+    context: &'ctx Context,
+    name: String,
```
rroelke, line 303: This is a different approach than most of the other Builders which allocate the thing right away and then call the underlying tiledb functions.  I see that it is necessary here because of the way the allocation function looks.  Maybe it would be nice to do this for the other Builders too though (not in this PR, if ever, obviously) - so all the setter methods succeed and then `build` does all the C API work.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        Builder {
+            context,
+            name: name.to_owned(),
+            dtype,
+            cell_val_num: 1,
+            ordered: false,
+        }
+    }
+
+    pub fn cell_val_num(self, cell_val_num: u32) -> Self {
```
rroelke, line 320: This has to be nonzero right? Maybe (in another PR) we should use the `Nonzero` struct here.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        }
+    }
+
+    pub fn ordered(self, ordered: bool) -> Self {
+        Self { ordered, ..self }
+    }
+
+    pub fn build<T>(
+        self,
+        data: &[T],
```
rroelke, line 340: Having the arguments here is certainly different than our usual but I see the appeal, not having to worry about a separate lifetime for the data that's passed to the constructor.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        offsets: Option<&[u64]>,
+    ) -> TileDBResult<Enumeration<'ctx>> {
+        let mut c_enmr: *mut ffi::tiledb_enumeration_t = out_ptr!();
+        let name_bytes = self.name.as_bytes();
+        let c_name = cstring!(name_bytes);
+        let c_dtype = self.dtype.capi_enum();
```
rroelke, line 346: Should we have a little bit of code that checks `T` against `self.dtype` before proceeding?

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        } else {
+            None
+        };
+
+        b.build(&self.data[..], offsets)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
rroelke, line 446: I'd suggest also a `test_eq` which confirms that changing any of the fields makes the equality false.

rroelke (changes requested):
Review is done!  I left a bunch of comments, the ones I think merit a follow-up I marked with :eyes: but these are all pretty small so it should be just a small quick step before approval.

This is all with the caveat that the ever-looming question of how to actually interface with data looms a lot larger over this PR. You mentioned arrow, that's probably a nice part of it, but probably not the whole story.  But I'm not requesting any changes of that sort here.

tiledb/api/src/array/enumeration/strategy.rs
```diff
@@ -0,0 +1,161 @@
...
+derive_wtf_cmp!(u32);
+derive_wtf_cmp!(u64);
+derive_wtf_cmp!(i8);
+derive_wtf_cmp!(i16);
+derive_wtf_cmp!(i32);
+derive_wtf_cmp!(i64);
+
+impl WtfNumCmp for f32 {
+    fn compare(&self, other: &Self) -> Ordering {
+        // Handle either is a NaN
```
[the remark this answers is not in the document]
davisp: For this particular case that shouldn't matter since the sort is followed immediately followed by dedup which will only leave a single NaN. And then beyond that, Enumeration values are internally treated as opaque bytes without interpreting so even having multiple NaN values with different bit level representations won't change the behavior.

tiledb/api/src/array/enumeration/mod.rs
davisp: The part I was being lazy on is the conversion of `Box<[u8]>` back into the datatype it came in as. Now that I have the basics figured out after fighting with the whole "fn_typed calls a generic function" I'll spend some extra time writing some conversion utilities and then some strategies for using those to generate data based on a datatype/cell_val_num combination. Only question is if you want me to do that as part of this PR or separate it out?

tiledb/api/src/array/enumeration/mod.rs
davisp: Oooh! Nice find. I don't know of any place where a cell_val_num of zero is a thing so that would work nicely.

tiledb/api/src/array/enumeration/mod.rs
davisp: Yeah, I haven't quite convinced myself whether or not adding the two lifetimes for data/offsets is worth it or not. I'm open to suggestion either direction.

tiledb/api/src/array/enumeration/mod.rs
davisp: Not a bad idea since we already have `Datatype::is_compatible_type` available. I'll add it.

tiledb/api/src/array/enumeration/strategy.rs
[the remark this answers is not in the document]
davisp: NaN is passable yeah. Inside enumerations we don't do any interpretation of the bytes given by the user. So long as they are the correct number of bytes for each value we just store them and spit them back out. So theoretically, multiple NaN's are even possible. So long as the bit patterns are distinct is all we care about internally.

TotalOrder looks like part of making this solution simpler. I keep forgetting about blanket trait implementations so that should just turn this into a much simpler trait.

tiledb/api/src/array/enumeration/mod.rs
rroelke: I think we can do a separate PR

tiledb/api/src/array/enumeration/mod.rs
rroelke: I created sc#44469

tiledb/api/src/array/enumeration/strategy.rs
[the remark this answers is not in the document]
rroelke: Oh, I meant my first paragraph as a lamentation, I think we speciifcally cannot use it because we need it for something other than floats, and the crate didn't give us that support.  Unless you want to call the implementation inside of what you've done here.

tiledb/api/src/array/enumeration/mod.rs
davisp: Huh. Fair point. Constructor args is actually easy enough since I should be able to do that with a single lifetime parameter. I'll give that a quick whirl and if it works out then great. But if the compiler yells too loudly I'll leave it as is.

tiledb/api/src/array/enumeration/mod.rs
davisp: Actually, this is kind of a bad idea. The original reason for providing this as a generic function was so that folks didn't have to do their own conversion to '&[u8]`. However, all of the enumeration strategies rely on being able to pass `'&[u8]` for any datatype. And that'll likely be the type passed whenever we provide an arrow translation API (based on my vague memory of the Array C interface).

tiledb/api/src/array/enumeration/strategy.rs
```diff
@@ -0,0 +1,161 @@
...
+use proptest::collection::vec;
+use proptest::prelude::*;
+
+use crate::array::EnumerationData;
+use crate::datatype::strategy::*;
+use crate::{fn_typed, Datatype};
+
+pub fn prop_enumeration_name() -> impl Strategy<Value = String> {
+    proptest::string::string_regex("[a-zA-Z0-9_]+")
+        .expect("Error creating attribute name strategy")
```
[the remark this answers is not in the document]
davisp: Fixed

tiledb/api/src/array/enumeration/strategy.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{Context, Factory};
+
+    /// Test that the arbitrary attribute construction always succeeds
+    #[test]
+    fn attribute_arbitrary() {
```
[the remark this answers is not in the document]
davisp: Fixed.

tiledb/api/src/array/enumeration/strategy.rs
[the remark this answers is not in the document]
davisp: Fixed. I linked this earlier, but I figure I should paste a link to the final implementation for posterity.

https://github.com/TileDB-Inc/tiledb-rs/blob/2d5c2940affe22ee36175f424bb5cb5bdccac9ba/tiledb/api/src/util/strategy.rs

tiledb/api/src/array/enumeration/strategy.rs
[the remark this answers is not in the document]
davisp: Fixed with the AnyNumCmp trait.

tiledb/api/src/array/enumeration/mod.rs
davisp: Will do.

tiledb/api/src/array/enumeration/mod.rs
davisp: Updated!

tiledb/api/src/array/enumeration/mod.rs
davisp: I don't think the two lifetimes is worth it, but I'll try the constructor version abiove.

tiledb/api/src/array/enumeration/mod.rs
davisp: Done.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+    pub fn new<T: 'static>(
+        context: &'ctx Context,
+        name: &str,
+        dtype: Datatype,
+        data: &'data [T],
+        offsets: Option<&'offsets [u64]>,
+    ) -> Self {
+        let data = unsafe {
+            std::slice::from_raw_parts(
+                data.as_ptr() as *const std::ffi::c_void as *const u8,
```
rroelke, line 275: This is definitely wrong if `T` is `String`, but I think we can put this in the usual bucket of "things we will worry about later".

I created sc#44537 and will drop a note about this there.

tiledb/api/src/array/enumeration/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+    }
+}
+
+pub struct Builder<'ctx, 'data, 'offsets> {
+    context: &'ctx Context,
+    name: String,
+    dtype: Datatype,
+    cell_val_num: u32,
+    ordered: bool,
+    data: &'data [u8],
```
rroelke, line 253: Nice!

tiledb/api/src/array/enumeration/strategy.rs
```diff
@@ -0,0 +1,95 @@
...
+    any::<bool>()
+}
+
+fn do_cmp<T: AnyNumCmp>(a: &T, b: &T) -> Ordering {
+    a.cmp(b)
+}
+
+fn prop_enumeration_values(datatype: Datatype) -> BoxedStrategy<Box<[u8]>> {
+    fn_typed!(datatype, DT, {
+        vec(any::<DT>(), 1..1024)
```
rroelke, line 30: Minor request to change these to named constants so we can grep more easily for them in a future where we want them to be configurable

tiledb/api/src/util/strategy.rs
rroelke, line 1: #43 adds a separate `tiledb_utils` crate, this might feel at home there

rroelke (approved):
+1 !

Nice.
I left one tweak comment but otherwise let's get this into `main`!

tiledb/api/src/array/enumeration/strategy.rs
davisp: Done.

tiledb/api/src/array/enumeration/mod.rs
davisp: Done. That function now only accepts the `u{8,16,32,64}`, `i{8,16,32,64}`, and `f{32,46}`

tiledb/api/src/util/strategy.rs
davisp: I saw that. I'll open a story to move it once #43 is merged.

PR #26, opened 2024-03-25
Dump libtiledb.so install log if installation fails

rroelke:


davisp:
Closing in favor of #30

PR #27, opened 2024-03-25
use pkg-config to find system libtildb

perrygeo:
:wave: Excited to see the rust bindings taking shape! To build against install prefixes outside of `/opt/tiledb`, the [pkg-config](https://docs.rs/pkg-config/latest/pkg_config/) package is one potential solution. It grabs all the system information at build time, tries to find `libtiledb > 2.4.0` (version TBD), and emits the appropriate flags. Working well locally but gh actions needs convincing.

TODO

* [ ] fix CI to install to a known prefix?
* [ ] minimum supported libtiledb version?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,1 @@
 [build]
-rustflags = ["-C", "link-args=-Wl,-rpath,/opt/tiledb/lib"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -701,6 +701,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3fdb12b2476b595f9358c5161aa467c2438859caa136dec86c26fdd2efe17b92"
 
 [[package]]
+name = "pkg-config"
+version = "0.3.30"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d231b230927b5e4ad203db57bbcbee2802f6bce620b1e4a9024a07d94e2907ec"
+
+[[package]]
 name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1030,6 +1036,9 @@ dependencies = [
 [[package]]
 name = "tiledb-sys"
 version = "0.1.0"
+dependencies = [
+ "pkg-config",
+]
 
 [[package]]
 name = "tiledb-test"
--- a/tiledb/sys/Cargo.toml
+++ b/tiledb/sys/Cargo.toml
@@ -2,3 +2,6 @@
 name = "tiledb-sys"
 version = { workspace = true }
 edition = { workspace = true }
+
+[build-dependencies]
+pkg-config = "0.3.30"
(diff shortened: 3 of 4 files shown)


davisp:
Howdy, @perrygeo!

This looks mostly reasonable and will likely be the basis of what we adopt. We've got an item on the agenda to address this but started of with the mindset of "what's the least amount of developer friction needed to get started" which we decided would be easiest with "just agree that we all have a version installed in `/opt/tiledb` and we'll wait until we have access to a Windows developer before dealing with that can of worms".

The base issue is developer friction in that it's a not uncommon requirement to have a not-system-installed library so that while doing work on the wrappers we can easily modify the libtiledb.so being used for debugging without making system level requirements. After seeing this and doing a bit of Googling, it looks like just inserting `/opt/tiledb` into the `PKG_CONFIG_PATH` environment variable will solve the `pkg-config` discovery part of things, but unfortunately there's one last issue.

It turns out, that Cargo is fairly opinionated on system library discovery, so our reliance on the `.cargo/config.toml` including the RPATH setting is currently somewhat crucial. I would love to figure out a way to make that dynamic based on the results of `pkg-config` or some other mechanism available via build.rs but near as I can tell, there's no standard way of doing this currently that'll work in all environments seamlessly.

Also, I'm open to any and all ideas you might have here. The general idea is to not leave new contributors in the lurch if a simple `cargo test` blows up for someone that just cloned this repository. One off the cuff idea would be to complain *VERY LOUDLY* if we add `/opt/tiledb` to the `PKG_CONFIG_PATH`, detect that we found a library there, and then not find `/opt/tiledb` in `RUSTCFLAGS` or something similar. Although, my initial reaction to that is also not super awesome as requiring users to modify a global Rust environment variable just for TileDB would be awkward to say the least.

Also as an aside, our current minimum version is just "most recent TileDB release". We'll obviously begin to extend that once we figure out how much of a pain it is to handle API changes. But I fully expect that we'll have a story before too long that will basically be summarized as "figure out how far back we can support" to define our minimum version support.

That's probably enough exposition for now. Feel free to throw anything at the CI to see if you can find a reasonable approach here, but keep in mind that for the moment at least we're more or less dedicated to treating "developer working on this repository" as the primary user story. So having some minimal amount of config required to get started is the primary goal. Long term we'll obviously switch to requiring "developers of this repository" having to do extra work for custom libtiledb.so installs, but for now its two or three developers vs zero users. Once we're a few developers vs thousands of installs that calculus will obviously change.

Thanks again for the initial PR. Feel free to reach out if you need any help or want to bounce any ideas off me.

perrygeo:
Thanks for the explanation. The "developer working on this repository" story is probably the right focus, which is largely what motivated this change. The assumption is that tiledb devs probably already have a libtiledb somewhere discoverable, why not use that?

I'm able to build and run the tests fine without any hardcoded configuration in `.cargo/config.toml`. I believe that pkg-config produces all the necessary flags based on the environment but I might be missing something.

Since this is somewhat unclear on how to proceed, I'll close this PR for the moment (it's a trivial amount of work to recreate) but will probably continue to use it locally for convenience.

PR #28, opened 2024-03-26
Finish implementing conversion between Arrow field and TileDB Attribute

rroelke:
Here we finish implementing the arrow <=> TileDB conversion for Attribute. This mostly required handling the fill values, plus some other incidentals.

`arrow_schema::Field` and `tiledb::Attribute` unsurprisingly do not have a fully overlapping set of fields.  But `Field` does have a generic metadata map.  In this pull request we add structures `AttributeMetadata` and so on, which we can use to serialize any extra fields of an `Attribute` into an `arrow_schema::Field`'s metadata map.  The tests show that we can use this to write a mostly-invertible* function.

("mostly" because I'm not sure we can carry *other* Field metadata into tiledb and then bring it back)

Resolves sc#43717.


Cargo.toml
```diff
@@ -8,6 +8,7 @@ rust-version = "1.72"
 version = "0.1.0"
 
 [workspace.dependencies]
+serde_json = { version = "1.0.114", features = ["float_roundtrip"] }
```
rroelke, line 11: `float_roundtrip` is a needed feature to preserve the precision of floats when serializing and deserializing.  Without this there were slight precision errors in the invertibility test.

tiledb/arrow/src/datatype.rs
```diff
@@ -130,10 +130,60 @@ pub fn is_same_physical_type(
...
+            Just(DT::Timestamp(TU::Microsecond, None)),
+            Just(DT::Timestamp(TU::Nanosecond, None)),
+            Just(DT::Time64(TU::Microsecond)),
+            Just(DT::Time64(TU::Nanosecond))
+        ]
+    }
+
+    pub fn arbitrary_arrow_implemented(
+    ) -> impl Strategy<Value = arrow_schema::DataType> {
+        tiledb_test::datatype::arbitrary_implemented().prop_map(|dt| arrow_type_physical(&dt).expect("Datatype claims to be implemented but does not have an arrow equivalent"))
```
davisp, line 165: How in the world is `cargo fmt` ok with this line?

tiledb/test/src/attribute.rs
```diff
@@ -12,23 +14,41 @@ pub fn arbitrary_name() -> impl Strategy<Value = String> {
...
+    context: &Context,
+    datatype: Datatype,
+) -> impl Strategy<Value = TileDBResult<Attribute<'_>>> {
+    proptest::prelude::any::<bool>()
+        .prop_flat_map(move |nullable| {
+            (Just(nullable),
+            if nullable { proptest::prelude::any::<bool>().bind() } else { Just(false).bind() })
+                .prop_flat_map(move |(nullable, fill_nullable)| {
+                    fn_typed!(arbitrary_fill_value, datatype => {
+                        (arbitrary_name(), Just(nullable), crate::filter::arbitrary_list_for_datatype(context, datatype), arbitrary_fill_value, Just(fill_nullable))
```
davisp, line 34: I know this is in a fn_typed! block so the formatter doesn't touch it. Could we pull out the arbitary_list_for_datatype into a temporary?

davisp (approved):
+1. As per usual, my only comments are style nits that you can ignore since they obviously pass `cargo fmt`.

tiledb/arrow/src/datatype.rs
rroelke: It's a pretty excellent question. I have no idea.  The formatter doesn't touch it if I manually tweak it so I'll just do that.

tiledb/test/src/attribute.rs
rroelke: Actually there's a different variant of `fn_typed` which does play well with the formatter, so I've changed this to use that instead.

PR #32, opened 2024-03-27
Update proptest strategies to produce a set of inputs to our APIs, instead of the structs directly

rroelke:
Prior to this pull request, our proptest strategies have generated our API structs directly - the Attribute, Dimension, Domain, Schema, etc.  This has led to a lot of difficulty as our strategies get more and more complex. Without being able to deeply clone an object we have no way to have a strategy depend on the contents of a Domain, for example, and we are reluctant to add a deep clone function.

This pull request modifies our strategies to instead produce a set of input values which can be used to construct our structures instead.  This way we can have no qualms about deep cloning, and we might even be able to benefit further from the new cloneable objects in the future.

There are other benefits briefly described in the shortcut story.

For each of our API objects, we add another companion structure which I have called `Data` - e.g. for `Attribute`, there is an `AttributeData`, and so on.  These companions implement `TryFrom` to be created from the thing they want to represent, and they implement a new trait `Factory` for creating objects of the thing they want to represent.  We use the former trait as a new way of implementing `Debug`, and we use the latter to create the structures from what we are given by proptest.

Finally, we see some payoff in the `tiledb_test::schema::arbitrary` strategy, which can now read the generated Domain in order to avoid choosing incompatible coordinate filters.

Resolves sc#43928.

### DIFF
--- a/tiledb/api/src/array/attribute.rs
+++ b/tiledb/api/src/array/attribute.rs
@@ -1,17 +1,19 @@
 extern crate tiledb_sys as ffi;
 
+use std::borrow::Borrow;
 use std::convert::From;
 use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::ops::Deref;
 
+use serde::{Deserialize, Serialize};
 use serde_json::json;
 
 use crate::context::Context;
 use crate::convert::{BitsEq, CAPIConverter};
 use crate::error::Error;
-use crate::filter_list::{FilterList, RawFilterList};
+use crate::filter_list::{FilterList, FilterListData, RawFilterList};
 use crate::fn_typed;
-use crate::{Datatype, Result as TileDBResult};
+use crate::{Datatype, Factory, Result as TileDBResult};
 
 pub(crate) enum RawAttribute {
     Owned(*mut ffi::tiledb_attribute_t),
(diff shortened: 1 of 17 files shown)


davisp (approved):
Everything looks awesome.

The only thing I'm not 100% on is whether the Data classes and Factor trait belong in the test crate instead of the API crate.

Although, if memory serves, there's a rule along the lines of not being able to implement a trait for a type in a different crate which I'd guess breaks the TryFrom implementations. Its not a huge deal either way and obviously can always be revisited if we decide that having those bits mixed into the API documentation is too annoying or something.

tiledb/api/src/array/attribute.rs
```diff
@@ -498,6 +478,85 @@ impl<'ctx> Builder<'ctx> {
...
+    pub nullability: Option<bool>,
+}
+
+/// Encapsulation of data needed to construct an Attribute
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct AttributeData {
+    pub name: String,
+    pub datatype: Datatype,
+    pub nullability: Option<bool>,
+    pub cell_val_num: Option<u32>,
```
davisp, line 494: Is the idea behind the Option's that we want to test when some of the builder methods are **not** called?

rroelke:
> Everything looks awesome.
> 
> The only thing I'm not 100% on is whether the Data classes and Factor trait belong in the test crate instead of the API crate.
> 
> Although, if memory serves, there's a rule along the lines of not being able to implement a trait for a type in a different crate which I'd guess breaks the TryFrom implementations. Its not a huge deal either way and obviously can always be revisited if we decide that having those bits mixed into the API documentation is too annoying or something.


I think it's a little weaker than that, you can implement traits for types defined in other crates - but the impl has to have no chance that the upstream crate could make a conflict.  I've definitely run into that a few times too and it sounded like you did too trying to make an impl for `T: AsRef<str>` on the enumerations PR.

So that's one reason to put these in this crate, but my decision was based on wanting to put them in the `Debug` implementations.  Testing is the primary reason we've done this, but not the only reason.  Using the JSON-ified thing in `Debug` requires them to be in this crate, and maybe future stuff we do could have a similar requirement.

(and if not I might even have said put them in yet another sub-crate rather than the testing one)

tiledb/api/src/array/attribute.rs
rroelke: It's not a required field for the Thing, so I decided to write that down here.  Similar to the decision for the Builder whether to have a field be a `new` argument or a setter method

PR #33, opened 2024-03-27
Wrap API and port stats example to Rust.

abigalekim:
I provide wrapper functions for the stats API in Rust, and also ported an example successfully to show that they work.


tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+       Ok(())
+    } else {
+       Err(Error::from("TileDB stats were not successfully reset."))
+    }
+}
+
+pub fn stats_dump_str() -> TileDBResult<String> {
+    let mut stats_dump_c_str = std::ptr::null_mut::<std::os::raw::c_char>();
+    let res = unsafe {
+        ffi::tiledb_stats_dump_str(
```
rroelke, line 43: Is it safe to call this when stats are not enabled?

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
+use crate::error::Error;
+use crate::Result as TileDBResult;
+
+pub fn stats_enable() -> TileDBResult<()> {
```
rroelke, line 4: I'll offer the disclaimer that I know absolutely nothing about what this is doing inside core.

I think this is a little too C-like.  Maybe it's not a safety or correctness issue but it does seem like a performance issue if you turned stats on and then bad control flow forgot to turn them off.  I'd probably want this function to give me a struct back, whose `impl Drop` turns the stats collection off.  And maybe this struct has a method that gives you the stats back, instead of a raw function.

That doesn't solve all the problems, of course - it's still not quite right if you put multiples of them in your scope - but it will be more user-friendly than a 1:1 mapping between C and Rust functions

davisp, line 4: I think this should probably be just `enable()` vs `stats_enable()`. Rust conventions for calling functions is to import the parent module which means if we have it as `stats_enable()` then it'd look something like:

```rust
use tiledb::stats;

fn foo() -> {
    stats::stats_enable();
}
```

instead of

```rust
use tiledb::stats;

fn foo() -> {
    stats::enable();
}
```

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    let c_ret : i32 = unsafe {
```
davisp, line 5: You shouldn't need a type specification here.

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    if c_ret == ffi::TILEDB_OK {
+       Ok(())
+    } else {
+        Err(Error::from("TileDB stats were not successfully enabled."))
+    }
+}
+
+pub fn stats_disable() -> TileDBResult<()> {
+    let c_ret : i32 = unsafe {
```
davisp, line 17: Same for removing the type specification.

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    let schema = tiledb::array::SchemaBuilder::new(
+        &tdb,
+        tiledb::array::ArrayType::Dense,
+        domain,
+    )?
+    .add_attribute(attribute_a)?
+    .build();
+
+    tiledb::Array::create(&tdb, ARRAY_NAME, schema)
```
rroelke, line 62: Does this return an error if the array already exists?

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+
+const ARRAY_NAME: &str = "stats_array";
+const ATTRIBUTE_NAME: &str = "a";
+
+pub fn create_array(row_tile_extent : u32, col_tile_extent : u32) -> TileDBResult<()> {
+    let tdb = tiledb::context::Context::new()?;
+    let config: Config = tiledb::config::Config::new()?;
+    let vfs: VFS = tiledb::vfs::VFS::new(&tdb, &config)?;
+
+    let is_cur_dir = vfs.is_dir(ARRAY_NAME)?;
```
rroelke, line 16: The `quickstart_dense` example checks if the array exists already instead of removing the directory every time - to me that seems like a better way to play.  Creating the array isn't really the important part of this example

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub fn stats_enable() -> TileDBResult<()> {
+    let c_ret : i32 = unsafe {
+        ffi::tiledb_stats_enable()
+    };
+
+    if c_ret == ffi::TILEDB_OK {
+       Ok(())
+    } else {
+        Err(Error::from("TileDB stats were not successfully enabled."))
```
davisp, line 12: Style nit, I have no idea on the grammar words, but I'd prefer something along the lines of "Failed to enable stats" something something passive voice something I think?

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+    tiledb::stats::stats_enable()?;
+    query.submit()?;
+    let stats : String = tiledb::stats::stats_dump_str()?;
+    println!("{}", &stats);
+    tiledb::stats::stats_disable().map(|_| ())
+}
+
+fn main() {
+    let _ = create_array(1, 12000);
+    let _ = write_array();
```
rroelke, line 115: This suppresses the warning but the warning is there for a good reason.  I think an `expect` like in `quickstart_dense` is fine but there does need to be a notification if anything fails.

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+use tiledb::{Array, Result as TileDBResult};
+use tiledb::config::Config;
+use tiledb::vfs::VFS;
+use tiledb::Datatype;
+
+
+const ARRAY_NAME: &str = "stats_array";
+const ATTRIBUTE_NAME: &str = "a";
+
+pub fn create_array(row_tile_extent : u32, col_tile_extent : u32) -> TileDBResult<()> {
```
rroelke, line 11: I see that the C example doesn't have them but I do think examples should be well-commented, at least to say what the array we're working on looks like.

(I see I did not follow my own advice in the other example.. I will fix that)

rroelke (changes requested):

tiledb/sys/src/stats.rs
```diff
@@ -0,0 +1,25 @@
+extern "C" {
+    #[doc = "Enable internal statistics gathering. \n @return `TILEDB_OK` for success and `TILEDB_ERR` for error."]
```
davisp, line 2: Remove all of the `#[doc = ..]` attributes. We'll need to re-document the Rust API in `tiledb/api` so these don't add anything other than add noise. The other examples including docs just haven't been updated to remove the attributes yet.

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    if c_ret == ffi::TILEDB_OK {
+       Ok(())
+    } else {
+       Err(Error::from("TileDB stats were not successfully reset."))
+    }
+}
+
+pub fn stats_dump_str() -> TileDBResult<String> {
```
davisp, line 40: Do you have any idea what the difference between `stats_dump_str` and `stats_dump_raw_str` is? Looking through the code paths in core I can't figure out the difference as they both appear to just end up calling the same function:

https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/stats/global_stats.cc#L69-L71
https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/stats/global_stats.cc#L82-L84

Unless you know of a difference I'd just wrap `ffi::tiledb_stats_dump_str` and delete the other function. Then you should rename this function to just `dump() -> TIleDBResult<String>`.

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    let mut stats_dump_c_str = std::ptr::null_mut::<std::os::raw::c_char>();
```
davisp, line 41: Style nit, just rename this to `c_str` since `stats_dump` is implied by the containing function.

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    let res = unsafe {
+        ffi::tiledb_stats_dump_str(
+            &mut stats_dump_c_str as *mut *mut std::os::raw::c_char,
+        )
+    };
+    if res == ffi::TILEDB_OK {
+        assert!(!stats_dump_c_str.is_null());
+        let stats_dump = unsafe { std::ffi::CStr::from_ptr(stats_dump_c_str) };
+        Ok(String::from(stats_dump.to_string_lossy()))
```
davisp, line 50: Vague style nit, you might try `to_string_lossy().into_owned()` for a slightly easier to read conversion.

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+       Err(Error::from("TileDB stats were not successfully reset."))
+    }
+}
+
+pub fn stats_dump_str() -> TileDBResult<String> {
+    let mut stats_dump_c_str = std::ptr::null_mut::<std::os::raw::c_char>();
+    let res = unsafe {
+        ffi::tiledb_stats_dump_str(
+            &mut stats_dump_c_str as *mut *mut std::os::raw::c_char,
```
davisp, line 44: Prefer to use the `std::ffi::c_char` alias here. They changed that at some recent-ish point so things are a bit inconsistent but we've mostly standardized on the `std::ffi::*` names.

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+        )
+    };
+    if res == ffi::TILEDB_OK {
+        assert!(!stats_dump_c_str.is_null());
+        let stats_dump = unsafe { std::ffi::CStr::from_ptr(stats_dump_c_str) };
```
davisp, line 49: I'm pretty sure we're leaking the string returned by core here. I think an easy way to clean this up would be to add a RawStatsString for the `*mut std::ffi::c_char` pointer that implements drop, and then we can probably wrap the `RawStatsString` -> `String` conversion using a TryFrom or something.

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+        let rows: tiledb::array::Dimension =
+            tiledb::array::DimensionBuilder::new::<u32>(
+                &tdb,
+                "row",
+                Datatype::UInt32,
+                &[1, 12000],
+                &row_tile_extent,
+            )?
+            .build();
+        let cols: tiledb::array::Dimension =
```
davisp, line 31: Minor nit to add a blank line,

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+    let query =
+        tiledb::QueryBuilder::new(&tdb, array, tiledb::QueryType::Write)?
+            .layout(tiledb::array::Layout::RowMajor)?
+            .dimension_buffer_typed(
+                ATTRIBUTE_NAME,
+                data.as_mut_slice(),
+            )?
+            .build();
+
+    query.submit().map(|_| ())
```
davisp, line 79: Is this `.map(|_| ())` actually doing something important, or should it just be `query.submit()?` ?

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+            .dimension_range_typed::<i32>(0, &[1, 3000])?
+            .add_subarray()?
+            .dimension_range_typed::<i32>(1, &[1, 12000])?
+            .build();
+
+    tiledb::stats::stats_enable()?;
+    query.submit()?;
+    let stats : String = tiledb::stats::stats_dump_str()?;
+    println!("{}", &stats);
+    tiledb::stats::stats_disable().map(|_| ())
```
davisp, line 110: Samesies for if `?` would suffice instead of the map no-op.

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+fn main() {
+    let _ = create_array(1, 12000);
+    let _ = write_array();
+    let _ = read_array();
```
davisp, line 116: I'm pretty sure you probably want something like:

```rust
fn main() -> TileDBResult<()> {
    create_array(1, 12000)?;
    write_array()?;
    read_array()?;
}
```

davisp (changes requested):
Pretty good overall. I think we're leaking the stats strings so that needs to be cleaned up. Other than that it was just style nits that we've not been great about writing down.

tiledb/api/src/stats.rs
davisp: The call is safe as long as we properly account for whether it returns nullptr or an empty string. Though I was skimming the implementation and we should really check whether there are corner cases on that.

tiledb/api/src/stats.rs
davisp: A slight disagreement, I'd add the ability to do what you're asking, but leave these APIs. Sometimes you really do just want to enable stats so that you can report periodically. So basically, keep this API shape as is, but add a `with_stats_enabled()` API that returns some struct/enum/whatever that has a drop trait that disables things.

Although, it suddenly occurs to me to ask what happens if a drop trait fails? In this case its probably super unlikely but it might be something to read up on. I'm guessing panic land but who knows.

tiledb/api/examples/using_tiledb_stats.rs
davisp: Yup.

tiledb/api/examples/using_tiledb_stats.rs
davisp: This is a pattern from core because there are lots of issues re-running the test suite so there's a lot of defensive "remove anything to reset state". Maybe we could add a thing that auto deletes arrays using drop traits? Though that's not something I'll ask of this PR.

tiledb/api/examples/using_tiledb_stats.rs
davisp: My vague style preference is to only expect things that might fail that are being tested, everything else that we're not testing should be `?` to indicate its super unexpected to fail.

tiledb/api/src/stats.rs
rroelke: If the overhead of stats colleciton is noticeable, and this is eventually running in a multi-user environment, then there is definitely a risk that user A will turn on stats collection and forget to turn it off, and user B will yell at TileDB support that their queries are suddenly slower but nothing changed.

If user A wants to have stats enabled for all their workloads then core must support having stats collected at a better granularity than global. And for our API, maybe user A would make this thing a const.  That is an option (I think), to make whatever the structure is const-constructible.

Since Rust doesn't really have exception handling the only way for anything to fail is to return an error or to panic, and since Drop doesn't have a return type it can only panic.  This particular case doesn't really seem panic appropriate to me (I think the worst-case scenario is more or less what I outlined above, and maybe some disk usage quota issues if stats does any logging) but logging it as an error would be wise.

I guess that means we should start thinking about logging.  Oh boy.  Not for this PR though

tiledb/api/examples/using_tiledb_stats.rs
rroelke: I added `Array::exists`, I think we should use it here.

tiledb/api/examples/using_tiledb_stats.rs
rroelke: One of the unsatisfying aspects of the `Result` and `?` operators is that you do miss a lot of context when you finally get to the end.  I generally agree that `expect` should be avoided inside a call stack, but `main` is at the bottom, there's nothing else left to catch and handle the error except for the system which can't possibly add any context.  I endorse `expect` in the `main` function because then if it does fail, then there is a clean message on the command line saying where to start, which you might not get from the default "`main` returned `Err`" handler.

tiledb/api/examples/using_tiledb_stats.rs
rroelke: #34 was merged to `main` so now I have followed my own advice :)

tiledb/api/src/stats.rs
rroelke: Isn't there `tiledb::string::String` for this purpose?

tiledb/api/examples/using_tiledb_stats.rs
rroelke: `query.submit()?;` might give an unused return value warning?

As I understood it there's only one query allowed to run on an open array at a time, so `query` consumes the array you pass it, and then my thinking was that `submit` should either return back `Self` if you can re-submit the query, or the array if not.  So in the current API at least there is a return value here.

tiledb/api/src/stats.rs
abigalekim: Yeah, I think maybe the existence of both APIs (returning an object that automatically disables stats when it goes out of scope (effectively RAII stats collection) and the normal API works well. I'll add that!

tiledb/api/examples/using_tiledb_stats.rs
davisp: Hurms. I see the awkward there. Feel free to ignore this for now @abigalekim. We can clean it up when we figure out that query.submit API better since its not yet settled.

tiledb/api/src/stats.rs
abigalekim: When I implemented this I assumed that the memory allocated from tiledb_stats_dump_str would be moved into the Rust string object and thus not need to be memory managed further. Although yeah it would be nice to implement a string wrapper type that drops as needed. I think I'll just implement a wrapper class that implements drop?

I think tiledb_string_t is a wrapper around the tiledb_string_t class, not necessarily for handling C to Rust string conversions. But I'm probably 75% sure of that.

tiledb/api/examples/using_tiledb_stats.rs
abigalekim: Hm, as for now I just switched it to a question mark and added Ok(()) at the bottom of each function. I think this should suffice for now and makes more sense given the intention of the error messages.

tiledb/api/examples/using_tiledb_stats.rs
abigalekim: Did some docs for the example!

tiledb/api/examples/using_tiledb_stats.rs
```diff
@@ -0,0 +1,143 @@
+extern crate tiledb;
+use tiledb::config::Config;
+use tiledb::vfs::VFS;
+use tiledb::Datatype;
+use tiledb::{Array, Result as TileDBResult};
+
+const ARRAY_NAME: &str = "stats_array";
+const ATTRIBUTE_NAME: &str = "a";
+
+/// Creates a dense array at URI `QUICKSTART_DENSE_ARRAY_URI()`.
```
rroelke, line 10: Different constant name for this one!

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,58 @@
...
+pub fn dump() -> TileDBResult<Option<String>> {
+    let mut c_str = std::ptr::null_mut::<std::ffi::c_char>();
+    let c_ret = unsafe {
+        ffi::tiledb_stats_dump_str(&mut c_str as *mut *mut std::ffi::c_char)
+    };
+    if c_ret == ffi::TILEDB_OK {
+        assert!(!c_str.is_null());
+        let stats_dump = unsafe { std::ffi::CStr::from_ptr(c_str) };
+        if stats_dump.is_empty() {
+            return Ok(None);
```
rroelke, line 43: I didn't see the full discussion about the empty string case - choosing not to free it is intentional, right?

Good use of the option type though.

rroelke (approved):
Deferring to Paul for approval but this looks good to me, thanks :)

tiledb/api/src/stats.rs
abigalekim: I think I'm supposed to free in the case of an empty string too--good catch!

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,85 @@
...
+    };
+    let datas: Vec<Metrics> =
+        serde_json::from_str::<Vec<Metrics>>(dump.as_str()).map_err(|e| {
+            Error::Deserialization(
+                format!("Failed to deserialize stats JSON value {}", dump),
+                anyhow!(e),
+            )
+        })?;
+    Ok(Some(datas))
+}
```
davisp, line 85: I think `if let` would make this quite a bit shorter. I haven't attempted compiling this, but something along these lines maybe?

```suggestion
pub fn dump_json() -> TileDBResult<Option<Vec<Metrics>>> {
    if let Some(dump) = dump()? {
        let data = serde_json::from_str::<Vec<Metrics>>(dump.as_str()).map_err(|e|  {
            ... same as you had ....
        })?;
        Ok(data)
    } else {
        None
    }
}
```

tiledb/sys/src/stats.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    //pub fn tiledb_stats_dump(out: *mut libc::FILE) -> i32;
+
+    pub fn tiledb_stats_dump_str(out: *mut *mut ::std::ffi::c_char) -> i32;
+
+    //pub fn tiledb_stats_raw_dump(out: *mut libc::FILE) -> i32;
+
+    pub fn tiledb_stats_raw_dump_str(out: *mut *mut ::std::ffi::c_char) -> i32;
+
+    pub fn tiledb_stats_free_str(out: *mut *mut ::std::ffi::c_char) -> i32;
```
davisp, line 16: Minor style suggestion. I'd add the note that we're intentionally not wrapping the commented out lines so folks don't think we've just not gotten to them.

```suggestion
    pub fn tiledb_stats_enable() -> i32;
    pub fn tiledb_stats_disable() -> i32;
    pub fn tiledb_stats_reset() -> i32;
    pub fn tiledb_stats_dump_str(out: *mut *mut ::std::ffi::c_char) -> i32;
    pub fn tiledb_stats_free_str(out: *mut *mut ::std::ffi::c_char) -> i32;

    // These are intentionally left unwrapped.
    //pub fn tiledb_stats_dump(out: *mut libc::FILE) -> i32;
    //pub fn tiledb_stats_raw_dump(out: *mut libc::FILE) -> i32;
    //pub fn tiledb_stats_raw_dump_str(out: *mut *mut ::std::ffi::c_char) -> i32;
```

tiledb/api/src/stats.rs
```diff
@@ -0,0 +1,85 @@
...
+    let mut c_str = std::ptr::null_mut::<std::ffi::c_char>();
+    let c_ret = unsafe {
+        ffi::tiledb_stats_dump_str(&mut c_str as *mut *mut std::ffi::c_char)
+    };
+    if c_ret == ffi::TILEDB_OK {
+        assert!(!c_str.is_null());
+        let stats_dump = unsafe { std::ffi::CStr::from_ptr(c_str) };
+        let stats_dump_rust_str = stats_dump.to_string_lossy().into_owned();
+        let free_c_ret = unsafe {
+            ffi::tiledb_stats_free_str(&mut c_str as *mut *mut std::ffi::c_char)
```
davisp, line 55: The only thing I'm not 100% certain about. Might be because I've still got C++ brain, but can we guarantee that we won't panic inbetween these lines and leak that pointer? Usually panic means end of the process, but apparently not always, so we could technically end up leaking the stats string if someone using tiledb-rs is catching panics. I think?

davisp (changes requested):
I'm not entirely sure about that pointer free without using some sort of RawThing wrapper like all of our other examples. Though I'll wait for @rroelke to voice an opinion there before making specific suggestions.

My other two comments were slight style nits that you are free to ignore.

One thing to add though, a couple of tests in the stats.rs module to exercise the APIs would be good coverage. You don't have to assert anything about the contents of the stats returned though. Just something as simple as assert dump and dump_json() return None when stats is not enabled, then assert they return Some(_) when stats is enabled, then disable stats and assert None again would be enough to just show that these APIs don't cause explosions. The only oddity I can think of is you might need to create something that pokes a stat after enabling.

tiledb/api/src/stats.rs
rroelke: There is apparently an "unwinding panic" which calls `drop` of things on the stack in the manner you would expect from a C++ thrown exception.  I wouldn't really expect that from `to_string_lossy()` unless maybe it runs out of memory. The doc makes no mention of a panic scenario.

I'd never say no to a struct that wrapped this thing and did teh `tiledb_stats_free_str` in a `drop` implementation though.

davisp: I’m fine leaving it as a follow up since it seems very unlikely to be an actual issue in practice. I’ll open an issue to remind me to file the story since I’m on mobile. Good to merge by me.

davisp (approved):

PR #34, opened 2024-03-28
Add comments to quickstart_dense.rs and also make Subarray::dimension_range_typed more flexible

rroelke:
In a different branch I added `DimensionKey` to use the dimension range functions with an index or name lookup. This adds the same facility to `Subarray`.

### DIFF
--- a/tiledb/api/examples/quickstart_dense.rs
+++ b/tiledb/api/examples/quickstart_dense.rs
@@ -6,6 +6,7 @@ use tiledb::Result as TileDBResult;
 const QUICKSTART_DENSE_ARRAY_URI: &str = "quickstart_dense_array";
 const QUICKSTART_ATTRIBUTE_NAME: &str = "a";
 
+/// Returns whether the example array already exists
 fn array_exists() -> bool {
     let tdb = match tiledb::context::Context::new() {
         Err(_) => return false,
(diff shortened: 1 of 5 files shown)


tiledb/api/examples/quickstart_dense.rs
```diff
@@ -6,6 +6,7 @@ use tiledb::Result as TileDBResult;
 const QUICKSTART_DENSE_ARRAY_URI: &str = "quickstart_dense_array";
 const QUICKSTART_ATTRIBUTE_NAME: &str = "a";
 
+/// Returns whether the example array already exists
 fn array_exists() -> bool {
```
davisp, line 10: We should probably nip this pattern in the bud rather than inherit it from core. I'm thinking a `tiledb::Array::exists(&ctx, "some_uri_thing")` or some such?

tiledb/api/examples/quickstart_dense.rs
```diff
@@ -18,6 +19,12 @@ fn array_exists() -> bool {
     )
 }
 
+/// Creates a dense array at URI `QUICKSTART_DENSE_ARRAY_URI`.
```
davisp, line 22: Something to think about here but a goal of the test suite in core is to try and move to a place where we can run all of these things against different URIs to prove out our VFS implementations. One nice thing about TileDB is that just changing an array URI from `file:///some/path/to/array_name` to `s3://bucketname/array_name` makes things Just Work &trade;

I'm absolutely not saying we should do this work as part of this PR, but given that we're just starting our test suite from scratch we should probably keep this use case top of mind.

davisp (approved):
+1

tiledb/api/examples/quickstart_dense.rs
ihnorton: 🙇

PR #35, opened 2024-03-28
[LEARNING] Just some work from me refactoring to learn

davisp:
This is just a quick draft PR to show others some of the refactoring I did while trying to learn proptest based on what @rroelke has already written. If we find parts that are good we can extract them for a separate real PR and eclude all of the stuff I changed on a whim.


tiledb/test/src/attribute.rs
```diff
@@ -1,78 +1,67 @@
...
-                                        nullability: Some(fill_nullable),
-                                    }),
-                                    filters,
-                                }
-                            },
-                        )
-                        .bind()
-                })
-            })
+    fn_typed!(datatype, DT, {
```
davisp, line 29: @rroelke This and the removal of the custom lifetime strategy are likely the only things you'll really want to consider when reviewing this PR. Everything else is basically just style changes I made while reading code.

rroelke: Yeah, removing that is fine, now that we no longer need the `&'ctx Context` we get everything we need from `boxed`

tiledb/test/src/util.rs
```diff
@@ -0,0 +1,9 @@
+use proptest::prelude::*;
+
+pub fn prop_field_name() -> impl Strategy<Value = String> {
+    proptest::string::string_regex("[a-zA-Z0-9_]*")
+        .expect("Error creating field name strategy")
+        .prop_filter("Field names may not begin with reserved prefix", |name| {
```
rroelke, line 6: Is this generally true, or just true for attributes?

rroelke:
All looks fine to me, with one exception - I don't think `Default` is really appropriate where you added it.

tiledb/test/src/util.rs
davisp: Pretty sure it applies to domain names as well but I'll double check to make sure.

davisp: I should also mention, `field_name` is the naming convention used in core to mean "attribute or dimension name" in case that wasn't obvious.

davisp: Huh. Looks like not actually. Although `__coords` has a special meaning but in a way that we don't want to exclude it. I'll swap that back to a prop_attribute_name and prop_dimension_name.

davisp:
Closing in favor of #37

PR #36, opened 2024-03-28
Better programmatic information for Error type, and ensure Sync/Send

rroelke:
This story is mostly about ensuring Sync and Send for our custom Error type.  To do that, we must remove the `*mut ffi::` raw pointer.  And then, since we're in the area, we can update our Error to have less reliance, a practice which is discouraged in the Rust documentation.

To that end, we change `Error` to be an `enum` for different error classes, and remove `From<String>`.

We add the dependency `thiserror` for use with our Error type, and add `anyhow` to generalize errors coming from other crates (such as with deserialization).

Resolves sc#44032.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1006,6 +1006,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "thiserror"
+version = "1.0.58"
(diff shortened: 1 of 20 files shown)


davisp (approved):
+1

PR #38, opened 2024-03-28
capi_return function, ContextBound trait, fix methods which do not return Result

rroelke:
This pull request does a few things which hopefully should improve stability of our API.

1) in a few cases the code wrote in the assumption that libtiledb would return `TILEDB_OK`.  We discussed and agreed not to do this, so this request fixes those.

2) we add the `Context::capi_return` function and move all C API return value checking in that function. There are more return codes than just `TILEDB_OK` and `TILEDB_ERR`, and this makes it substantially easier for us to accommodate those into possibly different kinds of Rust API errors.

3) we add the ContextBound trait which is just a cute way to give everything `fn context(&self) -> &'ctx Context`.  This is helpful for propagating the `capi_return` function described above, and also might be handy for users of the API.

Resolves sc#44161

### DIFF
--- a/tiledb/api/src/array/attribute.rs
+++ b/tiledb/api/src/array/attribute.rs
@@ -9,7 +9,7 @@ use anyhow::anyhow;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 
-use crate::context::Context;
+use crate::context::{CApiInterface, Context, ContextBound};
 use crate::convert::{BitsEq, CAPIConverter};
 use crate::error::{DatatypeErrorKind, Error};
 use crate::filter_list::{FilterList, FilterListData, RawFilterList};
@@ -42,6 +42,12 @@ pub struct Attribute<'ctx> {
     raw: RawAttribute,
 }
 
+impl<'ctx> ContextBound<'ctx> for Attribute<'ctx> {
+    fn context(&self) -> &'ctx Context {
+        self.context
+    }
+}
+
 impl<'ctx> Attribute<'ctx> {
     pub(crate) fn capi(&self) -> *mut ffi::tiledb_attribute_t {
         *self.raw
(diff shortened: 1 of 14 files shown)


tiledb/api/src/array/attribute.rs
```diff
@@ -54,28 +60,20 @@ impl<'ctx> Attribute<'ctx> {
...
         let mut c_name = std::ptr::null::<std::ffi::c_char>();
-        let res = unsafe {
+        self.capi_return(unsafe {
             ffi::tiledb_attribute_get_name(c_context, *self.raw, &mut c_name)
-        };
-        if res == ffi::TILEDB_OK {
-            let c_name = unsafe { std::ffi::CStr::from_ptr(c_name) };
-            Ok(String::from(c_name.to_string_lossy()))
-        } else {
-            Err(self.context.expect_last_error())
```
davisp, line 64: Oooooh.... pretty! Removing all the boilerplate like this is awesome.

tiledb/api/src/array/schema.rs
```diff
@@ -1055,17 +953,17 @@ mod tests {
 
         // capacity change
         {
-            let cmp = start_schema(base.array_type())
-                .capacity((base.capacity() + 1) * 2)
+            let cmp = start_schema(base.array_type().unwrap())
+                .capacity((base.capacity().unwrap() + 1) * 2)
```
davisp, line 957: Just a style fyi, I've found adding the `TileDBResult<()>` return value and just shoving an `Ok(())` as the last line of the test nicely allows for using `?` instead of having to type out `unwrap()` or `expect(...)`.

davisp (approved):
+1

Who doesn't love -300 line PRs that add consistency and improved error checking?

davisp:
Ope, I'm working on rebasing my Enumerations PR on the new Error stuff and I've noticed that we're going to need a variant of this work for the few APIs that don't have a context available.

tiledb/api/src/array/schema.rs
rroelke: I tried that for a bit, but what I like about this is that if the test fails it tells you the line number it failed on, whereas you don't get that with result.

Maybe if we put a stack trace in our error type :)

davisp: Ah, that's a fair point.

PR #39, opened 2024-03-29
Use "strategy" crate feature for exporting proptest strategies instead of separate test crate

rroelke:
Not much more to it than the title says.  But why bother?  This would let us implement traits on our structures using the strategies, most notably `proptest::arbitrary::Arbitrary`.  And it also seems like a better face to put on something that developers (possibly us, for tables) might even want to use.


davisp (approved):
+1 on the general principle but with a couple random comments.

First things first, I think "strategy" as the feature name is a bit awkwardly nonspecific. Perhaps something like "proptest_strategies"? I'm fine using "strategy" as the module name internally though.

For code organization, I think moving strategies closer to where the code under test lives. Closer to your Option 2, but I think with the expectation that for anything where the strategy code gets largeish we'll split it out into a `modname/{mod.rs, strategy.rs}` similar to how the test setup is done in the nom repository we looked at the other day.

The last thing that we might also want to think about is how to expose the strategies for use in testing. Perhaps make a `use tiledb::strategy::prelude::*` mod that reexports all the property definitions? Unfortunately that seems error prone/manual to do. Although over time it should settle down once the wrappers mature? Certainly not something that needs to be solved for this PR.

rroelke:
> +1 on the general principle but with a couple random comments.
> 
> First things first, I think "strategy" as the feature name is a bit awkwardly nonspecific. Perhaps something like "proptest_strategies"? I'm fine using "strategy" as the module name internally though.
> 

Fine by me. I tried `proptest` initially but unsurprisingly that didn't work since it was the same name as the crate we're using :)

> For code organization, I think moving strategies closer to where the code under test lives. Closer to your Option 2, but I think with the expectation that for anything where the strategy code gets largeish we'll split it out into a `modname/{mod.rs, strategy.rs}` similar to how the test setup is done in the nom repository we looked at the other day.
> 

Definitely, makes sense to me.

> The last thing that we might also want to think about is how to expose the strategies for use in testing. Perhaps make a `use tiledb::strategy::prelude::*` mod that reexports all the property definitions? Unfortunately that seems error prone/manual to do. Although over time it should settle down once the wrappers mature? Certainly not something that needs to be solved for this PR.

rroelke:
Module re-org is done, @davisp LMK if you want to have another look or just approve.  In most cases I did move `attribute.rs` into `attribute/mod.rs` (for example) and then move the strategies into `attribute/strategy.rs`, the exception being `datatype` whose strategies were pretty short.

davisp (approved):
+1

PR #41, opened 2024-04-01
Add test that our arbitrary schemata can actually create an array, and update strategies to make it so

rroelke:
Even though you can create a schema, that doesn't mean it can turn into an array!  This pull request found that out and then modifies our schema generation strategies so that our test does what we would like.

A lot of the complexity is driven by the constraints on the webp filter, and also by the proptest library guidance to avoid `proptest_filter`.  Fortunately our strategy already generates a Domain before it generates the attribute filters, so we can pass the Domain down to the filter pipeline generator and use it to determine what we're able to do.

The other main challenge is the dimension/attribute name uniqueness. Solving this with `proptest_filter` does not work at all because shrinking does not work on our types yet.  I started out trying to solve this by generating a set of names at the beginning and threading them through the dimension and attribute strategies, but this proved to be surprisingly complicated, so in the end we just make a pass at the end to de-duplicate the names.  It's not really the "correct" solution but it means we can Do The Thing, which seems like a fine compromise to make.

### DIFF
--- a/tiledb/api/src/array/schema/strategy.rs
+++ b/tiledb/api/src/array/schema/strategy.rs
@@ -125,7 +125,7 @@ fn prop_schema_for_domain(
                     {
                         let dimgen = crate::array::dimension::strategy::prop_dimension_name();
                         for dim in domain.dimension.iter_mut() {
-                            if !names.insert(dim.name.clone()) {
+                            while !names.insert(dim.name.clone()) {
                                 dim.name = dimgen
                                     .new_tree(&mut runner)
                                     .unwrap()
@@ -136,7 +136,7 @@ fn prop_schema_for_domain(
                     {
                         let attgen = crate::array::attribute::strategy::prop_attribute_name();
                         for attr in attributes.iter_mut() {
-                            if !names.insert(attr.name.clone()) {
+                            while !names.insert(attr.name.clone()) {
                                 attr.name = attgen
                                     .new_tree(&mut runner)
                                     .unwrap()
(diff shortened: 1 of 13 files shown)


tiledb/api/src/array/schema/strategy.rs
```diff
@@ -103,9 +108,47 @@ pub fn prop_schema_for_domain(
                 offsets_filters,
                 nullity_filters,
             )| {
+                /*
+                 * Update the set of dimension/attribute names to be unique.
+                 * This probably ought to be threaded into the domain/attribute strategies
+                 * so that they have unique names in all scenarios, but this is way
+                 * easier as long as we only care about the Schema in the end.
+                 */
```
davisp, line 116: Do you have a rough idea of what percentage of schemas had non-unique field names? Remember that one of things we can use proptest for is asserting when an error should happen. I.e., instead of unique-ifying names we could turn this into a "if we have duplicates, assert that the schema fails to build". Though if its a fairly high percentage of cases then this is also reasonable as well.

tiledb/api/src/filter/strategy.rs
```diff
@@ -1,13 +1,28 @@
...
 
+#[derive(Clone)]
+pub enum StrategyContext {
+    Domain(ArrayType, Rc<DomainData>),
+    SchemaCoordinates(Rc<DomainData>),
+}
+
+/// Defines requirements for what a generated filter must be able to accept
+#[derive(Clone, Default)]
+pub struct Requirements {
```
davisp, line 21: Good naming convention!

davisp (approved):
+1

tiledb/api/src/array/schema/strategy.rs
rroelke: I think it was pretty high because our name generator does allow zero and one-character names.

PR #43, opened 2024-04-03
Test which creates schema, creates array, loads schema, compares to original passes

rroelke:
This pull request contains a smorgasbord of things which all fall under the umbrella of enabling the test described in the title.

First and foremost, we look at sc#44377 - call `tiledb_array_schema_check` before constructing a schema.  There were a number of edge cases previously falling through our schema creation API which resulted in error during array creation or schema deserialization.  We should not be able to create an invalid schema, and calling this function in the `SchemaBuilder` gives us a lot more confidence that that is true.
- fixing where the webp filter appears, as it can only appear for uint8 attributes if the filter pipeline input is also uint8
- splitting Layout into CellOrder and TileOrder since not all values of Layout are valid for both contexts

The secondary concern which this PR addresses is that when we pass `None` to the C API, it fills in a default value.  If we put an object into the C API and then ask for it back, it has the default values filled in, which causes an equality comparison to fail.  Addressing this is what most of the changes are about:
- adding the `tiledb_utils` crate with trait `OptionSubset` to allow None to match against Some in comparisons
- making the Dimension filters be Option because if it is empty, then deserialization copies the schema's coordinates filters
- attaching `reinterpret_datatype` to the specific compression types which use it because deserialization effectively turned our Some into None, reversing the OptionSubset

There are also some other tweaks to the proptests.


tiledb/api/src/array/attribute/strategy.rs
```diff
@@ -117,7 +117,7 @@ pub fn prop_attribute(
         .map(|d| Just(d).boxed())
         .unwrap_or(prop_datatype_implemented().boxed());
 
-    datatype.prop_flat_map(move |datatype| {
+    datatype.prop_ind_flat_map(move |datatype| {
```
davisp, line 120: Can you write a few notes on why this change is required? Reading the docs all I'm seeing is that it creates an independent strategy that then can't be shrunk.

tiledb/api/src/array/domain/strategy.rs
```diff
@@ -67,10 +78,40 @@ mod tests {
...
+
+        let mut value = loop {
+            let value = strat.new_tree(&mut runner).unwrap();
+            if value.current().dimension.len() > 4 {
+                break value;
+            }
+        };
+
+        let init = value.current();
+        for _ in 0..1024 {
```
davisp, line 106: Is 1024 here just "a big number" so that we're guaranteed to reduce to the minimum number of dimensions? Could we just use `init.dimension.len()` or is it not guaranteed that simplify removes a dimension?

tiledb/api/Cargo.toml
```diff
@@ -15,10 +15,12 @@ proptest-derive = { version = "0.4.0", optional = true }
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = { workspace = true }
 thiserror = "1.0.58"
+tiledb-proc-macro = { workspace = true }
 tiledb-sys = { workspace = true }
+tiledb-utils = { workspace = true, features = ["serde_json"] }
 
 [dev-dependencies]
-tempdir = "0.3.7"
+tempfile = "3"
```
davisp, line 23: I'm vaguely curious on this change. Is it just that you needed a temporary file, and tmpfile is tempdir++?

tiledb/api/src/filter/strategy.rs
```diff
@@ -37,8 +37,98 @@ fn prop_bitwidthreduction() -> impl Strategy<Value = FilterData> {
...
+        .prop_filter(
+            "Input to delta filter cannot be floating-point type",
+            move |dt| dt_filter(*dt),
+        )
+        .prop_map(|dt| CompressionType::DoubleDelta {
+            reinterpret_datatype: Some(dt),
+        });
+
+    vec![delta.boxed(), double_delta.boxed()]
+    */
```
davisp, line 77: Are you needing these? Commented out code makes my eye twitch.

tiledb/api/src/filter/strategy.rs
```diff
@@ -277,16 +354,14 @@ fn prop_filter_pipeline_impl(
...
-                // already checked before being returned from
-                // prop_filter_for_datatype.
-                let next = filter
-                    .transform_datatype(&requirements.input_datatype.expect(
-                        "Input datatype required to construct pipeline",
-                    ))
-                    .unwrap();
+                // If the transformed datatype is None then we have a bug.
+                // Do not panic here, that will swallow what the pipeline looked like.
+                // Let the unit test will fail and print the input.
```
davisp, line 359: `s/test will fail/test fail/`

tiledb/proc-macro/src/option_subset.rs
```diff
@@ -0,0 +1,118 @@
...
+        impl #impl_generics OptionSubset for #name #ty_generics #where_clause {
+            fn option_subset(&self, other: &Self) -> bool {
+                #body
+            }
+        }
+    };
+    expanded.into()
+}
+
+fn option_subset_fields(fields: &syn::Fields) -> proc_macro2::TokenStream {
```
davisp, line 20: Super minor nit, but I prefer to sort functions by their call order. Currently we've got `expand -> option_subset_body -> option_subset_fields` as the call order, but the source code order is `expand -> open_subset_fields -> option_subset_body`.

tiledb/utils/src/option.rs
```diff
@@ -0,0 +1,795 @@
...
+        assert_not_option_subset!(
+            PointKDUnnamed(vec![0, 0]),
+            PointKDUnnamed(vec![0, 1])
+        );
+    }
+
+    #[test]
+    fn point_or_plane_named() {
+        #[derive(Debug, OptionSubset)]
+        struct PointOrPlane {
```
davisp, line 262: Clever name!

davisp (approved):
The only thing I'm not completely clear on is the `prop_ind_flat_map` specifically in terms of why its required and whether or not it has any other implications. Granted, that's a me problem, not something to change.

Great work!

davisp:
Just a note, I've opened SC#44542 to remind us to add enumeration support to the schema strategy once this lands.

tiledb/api/src/array/domain/strategy.rs
rroelke: Pretty much yep.  I'm not sure every simplify step will remove a dimension, so I picked a number.  But it would be better to read a value from `proptest::test_runner::Config::default()` since that is more representative of what our tests will actually do.

tiledb/api/src/array/attribute/strategy.rs
rroelke: As the tests were failing it was seeming like the schema wasn't really shrinking (the focus test of this PR would fail and print "minimum input is <thing>" with something that was very obviously not the minimum input).  So I was playing with shrinking for a bit and came to this as a way to simplify it a bit - we don't want to bother shrinking the datatype since the datatype is fixed.

(it turns out that it just doesn't shrink quickly enough and the default is only 1024 iterations, I did observe smaller schemata when I increased the iterations, but it doesn't change very much at each iteration so it goes very slowly. If we implement our own shrinking we probably ought to be more aggressive with shrinking cutting a lot, and complicating growing back a little)

I think at the time that I wrote this I misunderstood a bit what shrinking the datatype would actually mean.  But I think what happens under the hood In this particular case is as we try to simplify the attribute, one of the steps might find its way here and try to simplify the datatype; with `prop_ind_flat_map` it stops there, whereas with `prop_flat_map` we have a different ValueTree type. If we have `Just(dt)` then that doesn't do anything, if we have `prop_datatype_implemented()` which is basically "choose one of these things at random" then we also don't do anything.

So we save an indirect function call per simplification step, which is pretty much epsilon given that it also wants to re-run the whole test at each step.

Looks like all the `prop_ind_flat_map`s are ranging on a datatype strategy so in the end this doesn't really do anything meaningful.

tiledb/api/Cargo.toml
rroelke: If the assertion fails then there is a panic.  The old crate's `Tempfile` cleaned up in a regular `Drop` but not in a panic.  The new crate also cleans up if there is a panic.

tiledb/api/src/filter/strategy.rs
rroelke: Oh oops, I am also twitching!

PR #45, opened 2024-04-03
Add RawStatsString

davisp:
It was very unlikely that we would have paniced while this string was alive, but if we did and someone handles the panic, then the string would leak. Adding a RawStatsString wrapper for the pointer removes that edge case.

### DIFF
--- a/tiledb/api/src/stats.rs
+++ b/tiledb/api/src/stats.rs
@@ -1,10 +1,35 @@
 use std::collections::HashMap;
+use std::ops::Deref;
 
 use anyhow::anyhow;
 use serde::{Deserialize, Serialize};
 
 use crate::error::Error;
 use crate::Result as TileDBResult;
+
+pub(crate) enum RawStatsString {
+    Owned(*mut std::ffi::c_char),
+}
+
+impl Deref for RawStatsString {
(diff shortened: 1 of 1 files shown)


tiledb/api/src/stats.rs
```diff
@@ -1,11 +1,36 @@
...
+
+impl Drop for RawStatsString {
+    fn drop(&mut self) {
+        let RawStatsString::Owned(ref mut ffi) = *self;
+        let res = unsafe {
+            ffi::tiledb_stats_free_str(ffi as *mut *mut std::ffi::c_char)
+        };
+        // This is currently a hardcoded return value so this assertion only
+        // exists to know if that ever changes.
+        assert!(res == ffi::TILEDB_OK);
```
rroelke, line 30: `assert_eq` instead so that we see what the result is, if it does change?  Then we'd get a nicer bug report at least.

rroelke (approved):
I ran valgrind on our unit tests for funsies and there is a leak relating to stats.  Whether or not this fixes it it's a good change!

tiledb/api/src/stats.rs
davisp: Done!

PR #46, opened 2024-04-03
Groups API + Example

abigalekim:
Done with API minus metadata.


tiledb/api/src/array/mod.rs
```diff
@@ -34,6 +37,42 @@ impl Mode {
     }
 }
 
+impl TryFrom<ffi::tiledb_query_type_t> for Mode {
+    type Error = crate::error::Error;
+
+    fn try_from(value: ffi::tiledb_datatype_t) -> TileDBResult<Self> {
```
davisp, line 43: Pretty sure that should be `ffi::tiledb_query_type_t`

tiledb/sys/src/group.rs
```diff
@@ -0,0 +1,207 @@
+use crate::capi_enum::{tiledb_object_t, tiledb_query_type_t};
+use crate::tiledb_datatype_t;
+use crate::types::{
+    capi_return_t, tiledb_config_t, tiledb_ctx_t, tiledb_group_t,
+    tiledb_string_t,
+};
+
+extern "C" {
+    #[doc = " Creates a new TileDB group.\n\n **Example:**\n\n @code{.c}\n tiledb_group_create(ctx, \"my_group\");\n @endcode\n\n @param ctx The TileDB context.\n @param group_uri The group URI.\n @return `TILEDB_OK` for success and `TILEDB_ERR` for error."]
```
davisp, line 9: Please remove the all the doc strings and then make sure the first two functions are `tiledb_group_alloc` followed by `tiledb_group_free`.

tiledb/sys/src/types.rs
```diff
@@ -104,3 +104,9 @@ pub struct tiledb_vfs_t {
 pub struct tiledb_vfs_fh_t {
     _unused: [u8; 0],
 }
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct tiledb_group_t {
```
davisp, line 110: Lets keep these alphabetized.

tiledb/api/src/array/mod.rs
```diff
@@ -34,6 +37,42 @@ impl Mode {
...
+impl Debug for Mode {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        <Self as Display>::fmt(self, f)
+    }
+}
+
+impl Display for Mode {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        let mode_str = match *self {
+            Mode::Delete => String::from("Mode::Delete"),
```
rroelke, line 67: The compiler will generate this for you if you do `#[derive(Debug)]` and then defer to the debug format here (for a simple enum there's not really a meaningful difference)

tiledb/api/src/context.rs
```diff
@@ -1,15 +1,48 @@
...
+                    ObjectTypeErrorKind::InvalidDiscriminant(value as u64),
+                ))
+            }
+        })
+    }
+}
+
+impl Debug for ObjectType {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        <Self as Display>::fmt(self, f)
```
rroelke, line 32: Hm, I'd usually expect this to go the other way around, with `#[derive(Debug)]` and then having Display defer to the Debug format.

tiledb/api/src/group.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub fn open<S>(
+        context: &'ctx Context,
+        uri: S,
+        query_type: QueryType,
+    ) -> TileDBResult<Self>
+    where
+        S: AsRef<str>,
+    {
+        let ctx = context.capi();
+        let mut group_raw: *mut ffi::tiledb_group_t = std::ptr::null_mut();
```
rroelke, line 79: Use the `out_ptr!()` macro for the value we expect to be initialized by the C api

tiledb/api/src/group.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        let c_uri = cstring!(uri.as_ref());
+
+        context.capi_return(unsafe {
+            ffi::tiledb_group_alloc(ctx, c_uri.as_ptr(), &mut group_raw)
+        })?;
+
+        let query_type_raw = query_type.capi_enum();
+        context.capi_return(unsafe {
+            ffi::tiledb_group_open(ctx, group_raw, query_type_raw)
```
rroelke, line 89: If this returns `Err` then `group_raw` will be leaked

tiledb/api/src/group.rs
```diff
@@ -0,0 +1,386 @@
...
+            ffi::tiledb_group_get_query_type(
+                c_context,
+                Self::capi(self),
+                &mut c_type,
+            )
+        })?;
+        QueryType::try_from(c_type)
+    }
+
+    pub fn delete_group<S>(
```
rroelke, line 120: From reading the doc it sounds like what this does is "remove an entry from the group" - curious choice to name it `delete_group`.

Anyway, if we're collectively comfortable making changes to jargon like that, then I would lean towards making that swap... though obviously I acknowledge the strong case not to do this.

In general we haven't been very good about putting docstrings on our methods, I think we probably ought to start doing a better job of it so that we don't have to go back and write them all before a hypothetical release of the rust API...

Anyway, the docstring for the C function does say that the group must be opened in `MODIFY_EXCLUSIVE` mode, and AFAIK it would be fairly idiomatic Rust to embed a proof of that ownership in the type system.  Sounds like a fun nice-to-have.

rroelke, line 120: Reading a bit more stuff, I now believe that this method fully deletes the entire group - is that correct?

If so then none of the methods would be usable afterwards, so this ought to have a consuming `self` rather than `&self`

tiledb/api/src/group.rs
```diff
@@ -0,0 +1,386 @@
...
+                c_uri.as_ptr(),
+                c_relative,
+                c_name.as_ptr(),
+            )
+        })?;
+        Ok(())
+    }
+
+    pub fn delete_member<S>(
+        &self,
```
rroelke, line 171: Mutating methods should require `&mut self`

tiledb/api/src/group.rs
```diff
@@ -0,0 +1,386 @@
...
+        })?;
+
+        let query_type_raw = query_type.capi_enum();
+        context.capi_return(unsafe {
+            ffi::tiledb_group_open(ctx, group_raw, query_type_raw)
+        })?;
+        Ok(Self::new(context, RawGroup::new(group_raw)))
+    }
+
+    pub fn uri(&self, context: &'ctx Context) -> TileDBResult<String> {
```
rroelke, line 94: The context argument should just be `self.context()` right?  Seems like you can remove it from quite a few of these methods

rroelke (changes requested):
I'm not done reviewing but figured I'd publish my first set of comments in case you're still working today

tiledb/api/src/group.rs
davisp: I have some CI build stuff queued up and one thing I’ve been mulling over is adding a required CI check to start requiring new public APIs to be covered by docs. That means we’ll be on the hook for going back and documenting all the things we’ve already written, but it would at least mean going forward it’ll require maintaining docs.

I have absolutely no idea if it’s even possible. Or how to best make the transition, but does that seem like a reasonable approach to y’all?

davisp: I :heart: Rust.

https://doc.rust-lang.org/rustdoc/lints.html

tiledb/api/src/array/mod.rs
rroelke: I see you resolved this, have you pushed changes to update it? Or did you want to keep it the way it is?
Either is fine with me. Though I'm pretty sure this is exactly what `#[derive(Debug)]` will give you.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+    {
+        let ctx = context.capi();
+        let mut group_raw: *mut ffi::tiledb_group_t = std::ptr::null_mut();
+
+        let c_uri = cstring!(uri.as_ref());
+
+        context.capi_return(unsafe {
+            ffi::tiledb_group_alloc(ctx, c_uri.as_ptr(), &mut group_raw)
+        })?;
+
```
rroelke, line 88: If `tiledb_group_open` fails for some reason, then `group_raw` will leak. Make sure to construct the `RawGroup` as soon as you have a valid group handle

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+        unsafe { ffi::tiledb_group_free(&mut self.ffi) }
+    }
+}
+
+pub struct Group<'ctx> {
+    context: &'ctx Context,
+    raw: RawGroup,
+}
+
+impl<'ctx> ContextBound<'ctx> for Group<'ctx> {
```
rroelke, line 47: When you update your branch with `main` you can `#[derive(ContextBound)]` instead

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+        })?;
+
+        let query_type_raw = query_type.capi_enum();
+        context.capi_return(unsafe {
+            ffi::tiledb_group_open(ctx, group_raw, query_type_raw)
+        })?;
+        Ok(Self::new(context, RawGroup::new(group_raw)))
+    }
+
+    pub fn uri(&self, context: &'ctx Context) -> TileDBResult<String> {
```
rroelke, line 96: Methods with `self` shouldn't need the `context` argument since it is a field.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+            ffi::tiledb_group_get_query_type(
+                c_context,
+                Self::capi(self),
+                &mut c_type,
+            )
+        })?;
+        QueryType::try_from(c_type)
+    }
+
+    pub fn delete_group<S>(
```
rroelke, line 122: I really want to see a doc comment here which explains the difference between this and `delete_member`, I'm looking at the C docstrings in `generated.rs` and I'm honestly still unclear on it.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+        &self,
```
rroelke, line 123: Having a shared reference for this method, which the C docs says must be open in `MODIFY_EXCLUSIVE` mode, doesn't really seem right... an exclusive reference `&mut self` seems like the minimum of what should happen.  I think the heavier version would probably be to have separate types for the different modes, e.g. `ModifyExclusiveGroup` is the only one that can do this.  That's the direction the query API is heading.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+            ffi::tiledb_group_remove_member(
+                ctx,
+                Self::capi(self),
+                c_name_or_uri.as_ptr(),
+            )
+        })?;
+        Ok(())
+    }
+
+    pub fn get_member_count(
```
rroelke, line 195: Chatting with Paul in another PR, I think we're leaning towards `num_things` for the items which are collections.  `num_attributes`, `num_dimensions`... and here `num_members`.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+            ffi::tiledb_group_get_member_count(
+                ctx,
+                Self::capi(self),
+                &mut c_count,
+            )
+        })?;
+        Ok(c_count)
+    }
+
+    pub fn get_member_by_index(
```
rroelke, line 211: On `main` we now have a `DimensionKey` which will soon be re-named `LookupKey` by the fragment info PR. Using that, you can/should consolidate `get_member_by_index` and `get_member_by_name` into a single function `member`

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+                index,
+                &mut tiledb_uri as *mut *mut ffi::tiledb_string_t,
+                &mut tiledb_type,
+                &mut tiledb_name as *mut *mut ffi::tiledb_string_t,
+            )
+        })?;
+        let uri = TDBString {
+            raw: RawTDBString::Owned(tiledb_uri),
+        }
+        .to_string()?;
```
rroelke, line 233: The `?` is a little too early here, if it does do the early return then `tiledb_name` would be leaked.  I'd suggest
```
let uri = TDBString { raw: RawTDBString::Owned(tiledb_uri) }.to_string();
let name = TDBString { raw: RawTDBString::Owned(tiledb_name) }.to_string();
let uri = uri?;
let name = name?;
```
or something along those lines

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+        Ok(c_count)
+    }
+
+    pub fn get_member_by_index(
+        &self,
+        context: &'ctx Context,
+        index: u64,
+    ) -> TileDBResult<GroupInfo> {
+        let ctx = context.capi();
+        let mut tiledb_uri = std::ptr::null_mut::<ffi::tiledb_string_t>();
```
rroelke, line 217: Use `out_ptr!()` for these

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+
+    pub fn get_member_by_name<S>(
+        &self,
+        context: &'ctx Context,
+        name: S,
+    ) -> TileDBResult<GroupInfo>
+    where
+        S: AsRef<str>,
+    {
+        let ctx = context.capi();
```
rroelke, line 256: Most places this is named `c_context`, I think it's a good convention to follow since the name tells us immediately that it's something we will pass to an `ffi` function.

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+    };
+}
+
+impl Value {
+    pub fn c_vec(&self) -> (*const std::ffi::c_void, usize) {
+        value_typed!(self, _DT, vec, get_value_vec(vec))
+    }
+}
+
+trait ValueType {
```
rroelke, line 79: Looks like `From<Vec<T>>`, that would be more idiomatic

rroelke (changes requested):
Have to jump for the moment, I did not finish reviewing but will aim to do so tomorrow before our sync.

davisp:
Something has gone wonky with the history of this PR. I know things are moving fast so its not totally surprising. I'll spend some time tomorrow figuring out how to get @abigalekim's work rebased on top of main tomorrow since it looks like a lot of the issues are in support/infra code I've written.

tiledb/api/src/group/mod.rs
davisp:
```suggestion
    pub fn get_member_by_index(
        &self,
        context: &'ctx Context,
        index: u64,
    ) -> TileDBResult<GroupInfo> {
        let mut c_uri: *mut ffi::tiledb_string_t = out_ptr!();
        let mut c_name: *mut ffi::tiledb_string_t = out_ptr!();
        let mut c_type: ffi::tiledb_object_t = out_ptr!();
        context.capi_return(unsafe {
            ffi::tiledb_group_get_member_by_index_v2(
                context.capi(),
                Self::capi(self),
                index,
                &mut c_uri as *mut *mut ffi::tiledb_string_t,
                &mut c_type,
                &mut c_name as *mut *mut ffi::tiledb_string_t,
            )
        })?;

        let (uri, name) = {
            let uri = TDBString {
                raw: RawTDBString::Owned(c_uri),
            };
            let name = TDBString {
                raw: RawTDBString::Owned(c_name),
            };
            (uri, name)
        };

        let (uri, name) = (uri.to_string()?, name.to_string()?);
        let object_type = ObjectType::try_from(c_type)?;
        Ok(GroupInfo {
            uri,
            group_type: object_type,
            name,
        })
    }
```

davisp: @rroelke Has the `let thing = { ... some complicated stuff ... }` pattern in his query PR and I rather like that approach.

tiledb/api/src/group.rs
abigalekim: I think I'll add a TODO in the code for now and modify this once i finish the other code comments here.

tiledb/api/src/group/mod.rs
abigalekim: I thought that this was supposed to be a consuming reference?

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+
+    pub fn set_config(
+        &self,
+        context: &'ctx Context,
+        config: &Config,
+    ) -> TileDBResult<()> {
+        let ctx = context.capi();
+        let cfg = config.capi();
+        context.capi_return(unsafe {
+            ffi::tiledb_group_set_config(ctx, Self::capi(self), cfg)
```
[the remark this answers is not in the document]
abigalekim: No I don't think this is required.

PR #49, opened 2024-04-04
Initial api-coverage implementation

davisp:
A bit of an experiment that seems to actually be useful after a few hours of work. The basic idea is that this parses the generated.rs bindgen definitions, compares it to what we declare in `tiledb/sys`, and then also counts how many of those are actually called in `tiledb/api`.

Some example output: https://gist.github.com/davisp/25d5be6297d477238563bb313644b050


tools/api-coverage/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "api-coverage"
+edition.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
rroelke, line 7: Probably should remove the default comment

tools/api-coverage/src/api.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl<'ast> Visit<'ast> for APIVisitor {
+    fn visit_expr_call(&mut self, node: &'ast syn::ExprCall) {
+        if let syn::Expr::Path(path) = node.func.as_ref() {
+            let mut fnname = FnNameVisitor::default();
+            visit::visit_expr_path(&mut fnname, path);
+            let fnname = fnname.segments.join("::");
+            self.calls.insert(fnname);
+        } else {
+            println!("Weird guy here!: {:?}", node.func);
```
rroelke, line 35: What does this case mean? You expect it to be unreachable?

tools/api-coverage/src/main.rs
```diff
@@ -0,0 +1,207 @@
...
+        let mut uncalled: Vec<String> = Vec::new();
+        let mut unwrapped: Vec<String> = Vec::new();
+        let mut called = 0;
+        let mut declared = 0;
+        let mut generated = 0;
+        let mut mismatch = false;
+
+        for name in signature_names {
+            generated += 1;
+            let sig = self.generated_defs.signatures.get(&name).unwrap();
```
rroelke, line 144: Minor but you can avoid the extra hash lookup by doing `self.generated_defs.signatures.iter()` above and collecting into `Vec<(String, String)>` which I am pretty sure will sort on the key first.

tools/api-coverage/src/main.rs
```diff
@@ -0,0 +1,207 @@
...
+            let mut wrapped = false;
+            let sys_def = self.sys_defs.signatures.get(&name);
+            if sys_def.is_none() {
+                unwrapped.push(name.clone());
+            } else if sys_def.unwrap().inputs == sig.inputs {
+                wrapped = true;
+                declared += 1;
+            } else {
+                mismatch = true;
+                println!("SIGNATURE MISMATCH: {}", name);
```
rroelke, line 159: Should this, and the other mismatches, go to stderr?

tools/api-coverage/src/main.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        for name in signature_names {
+            generated += 1;
+            let sig = self.generated_defs.signatures.get(&name).unwrap();
+            if let Some(ignored) = self.ignored_defs.signatures.get(&name) {
+                assert_eq!(sig.inputs, ignored.inputs);
+                continue;
+            }
+
+            let mut wrapped = false;
```
rroelke, line 150: Understanding check:
We iterate over everything declared in `generated.rs`.
The paragraph above this filters out functions in the "ignored" list.
This paragraph checks if there is a file in the `sys` crate which matches the function we're looking at from `generated.rs`.
The paragraph below this checks if we call this function in the `api` crate with the `ffi` prefix.

All the while you accumulate (ngenerated, ndeclared, ncalled) so we can get the percentages of functions which we made a proper declaration for and a percentage of functions which we are thus far actually using.

rroelke (approved):
This is new turf for me so I mostly am here to ask a few questions.  Definitely a better way to compute our prorgress, and I bet it was fun to play with :)

tools/api-coverage/src/api.rs
davisp: That was just my in-progress learning of the AST. I'll make it a panic now that I've shown we don't have any thing in our code base that doesn't have a PathExpr for the function invocation expression.

tools/api-coverage/src/main.rs
davisp: I see your point, but sig is a `struct syn::Signature` so that'll probably mean sort_on with a lambda, but should be easy enough.

tools/api-coverage/src/main.rs
davisp: I don't think so. My main intention for going this far is to eventually have this be a required check in CI. The mismatch case is to specifically catch when a C API function's arguments change. As such, that means we'll want the report to be consistent which writing to separate streams can mess up.

davisp: Also for the CI angle, I also intend to exit non-zero after we've reached 100% on everything and then suddenly we detect less than 100% as that'll automatically flag any changes to the C API for us going forward.

tools/api-coverage/Cargo.toml
davisp: Fixed.

tools/api-coverage/src/api.rs
davisp: Fixed.

tools/api-coverage/src/main.rs
davisp: Fixed.

rroelke (approved):

davisp:
This PR is for [sc-45290]

PR #52, opened 2024-04-08
Wrapping fragment info APIs

davisp:
This is an initial wrapping of the FragmentInfo APIs. For the most part these are all fairly straight forward. However, I'm uncertain if the NonEmptyDomain and Range implementations are the best or if we couldn't do better there.

Also, there's not a lot of testing since there's no way to create a FragmentInfo without reading an array. I have ported the `fragment_info.cc` example  though to at least demonstrate the basics.


tiledb/api/src/non_empty_domain.rs
```diff
@@ -0,0 +1,111 @@
...
+    UInt16Range(u16, u16),
+    UInt32Range(u32, u32),
+    UInt64Range(u64, u64),
+    Int8Range(i8, i8),
+    Int16Range(i16, i16),
+    Int32Range(i32, i32),
+    Int64Range(i64, i64),
+    Float32Range(f32, f32),
+    Float64Range(f64, f64),
+    VarSizeRange(Box<[u8]>, Box<[u8]>),
```
rroelke, line 20: Is this meant to be attached to a specific datatype?
From what I've come to understand the var size unit could be any of the primitive types, right?

tiledb/api/src/non_empty_domain.rs
```diff
@@ -0,0 +1,111 @@
+use std::fmt::{Debug, Formatter, Result as FmtResult};
+
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+
+use crate::datatype::Datatype;
+
+#[derive(Serialize, Deserialize)]
```
rroelke, line 8: Probably add `Clone` to the derive lists, can't hurt to have and might be useful to somebody

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,667 @@
+use std::ops::Deref;
+
+use anyhow::anyhow;
+// use serde::{Deserialize, Serialize};
```
rroelke, line 4: Oops

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,667 @@
...
+                &mut c_config,
+            )
+        })?;
+
+        Ok(Config::from_raw(RawConfig::Owned(c_config)))
+    }
+
+    pub fn load(&self) -> TileDBResult<()> {
+        self.capi_return(unsafe {
+            ffi::tiledb_fragment_info_load(self.context.capi(), self.capi())
```
rroelke, line 70: Is this a required call for the fragment info to be useful?
If so it should be called in the build or new methods

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,667 @@
...
+    raw: RawFragmentInfo,
+}
+
+impl<'ctx> ContextBound<'ctx> for FragmentInfo<'ctx> {
+    fn context(&self) -> &'ctx Context {
+        self.context
+    }
+}
+
+impl<'ctx> FragmentInfo<'ctx> {
```
rroelke, line 50: Lots of the methods take a fragment index.
This is really a list of fragments with some extra metadata on top, right?
What do you think about embedding the elements in another structure which binds the index?
And then this parent has `fn fragment(idx: usize) -> TileDBResult<Fragment>`

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,667 @@
...
+
+    pub fn load(&self) -> TileDBResult<()> {
+        self.capi_return(unsafe {
+            ffi::tiledb_fragment_info_load(self.context.capi(), self.capi())
+        })?;
+
+        Ok(())
+    }
+
+    pub fn num_fragments(&self) -> TileDBResult<u32> {
```
rroelke, line 76: Our "has a list of stuff" structures are not consistent with the naming convention for getting the number of elements.
FilterList has `get_num_filters`
Domain has `ndim`
Schema has `nattributes`

We probably should make them all consistent.  I'm not sure if there's a convention for this, collections all use `len` which doesn't seem appropriate here.  I'd favor changing them all to `num_thing` or `nthing`

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,667 @@
...
+                self.capi(),
+                frag_idx,
+                &mut num,
+            )
+        })?;
+
+        Ok(num)
+    }
+
+    pub fn mbr<T: Into<LookupKey> + Clone>(
```
rroelke, line 336: What does "mbr" mean?
Good time to start adding some method documentation :)

tiledb/api/src/error.rs
```diff
@@ -28,6 +28,8 @@ impl Drop for RawError {
 
 #[derive(Clone, Debug)]
 pub enum DatatypeErrorKind {
+    ExpectedFixedSize,
+    ExpectedVarSize,
```
rroelke, line 32: Add some fields for context? The naem of the attribute would be helpful if it is available, but definitely at least it could have the cell val num that was actually found

tiledb/api/src/string.rs
```diff
@@ -29,6 +29,10 @@ pub struct TDBString {
 }
 
 impl TDBString {
+    pub(crate) fn from_raw(raw: RawTDBString) -> Self {
```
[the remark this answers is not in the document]
davisp: No idea.

tiledb/api/src/non_empty_domain.rs
davisp: Correct. I was just being lazy. I'll go back and add `Var${Type}Range` and do it properly now that I have the basics done.

tiledb/api/src/array/fragment_info.rs
davisp: I waffled on this a bit because this can be a relatively heavy operation depending on the array. I'll go peek at what Python does for a second opinion.

tiledb/api/src/array/fragment_info.rs
davisp: I waffled on this as well. My general preference is to err on the side of transparent wrappers because I've been bitten too many times by wrapper abstractions that prevent legitimate uses of an API. Though I think in this particular case you've convinced me to at least give the convenience approach a whirl.

tiledb/api/src/array/fragment_info.rs
davisp: FilterList was from before I really internalized the builder pattern, hence the `get_` prefix. My personal preference is for `num_${thing}` but its a lightly held preference. Totes agree that we should normalize one way or the other. I opened [sc-45388] to track the normalization work.

tiledb/api/src/array/fragment_info.rs
davisp: Totes fair. For reference, mbr stands for [minimum bounding rectangle](https://en.wikipedia.org/wiki/Minimum_bounding_rectangle). Its from geospatial stuff. In our case it comes from the RTree used by sparse arrays for dimension indexing.

tiledb/api/src/error.rs
davisp: Fair point. Will do.

tiledb/api/src/non_empty_domain.rs
davisp: Fixed.

tiledb/api/src/non_empty_domain.rs
davisp: Done.

tiledb/api/src/array/fragment_info.rs
davisp: Good catch, fixed.

tiledb/api/src/array/fragment_info.rs
davisp: I ended up having `build()` automatically call `load()` and then added an extra `build_without_loading` just in case we end up needing it for some reason. I could see in async land that we might want to create the FragmentInfoList in one thread and then shunt it off to a thread pool to do the I/O or something similar.

tiledb/api/src/array/fragment_info.rs
davisp: I ended up doing the following:

1. Rename FragmentInfo -> FragmentInfoInternal and made it a private type.
2. FragmentInfo is now a struct holds references to FragmentInfoInternal and an index
3. FragmentInfoList now has the non-index methods and can return specific fragments or iterate over all of them.

tiledb/api/src/array/fragment_info.rs
davisp: With the convenience stuff I ended up creating a new type named `MinimumBoundingRectangle` for this.

tiledb/api/src/error.rs
davisp: Turns out I didn't even end up using these so I've removed them. I keep wanting them, but I've also been figuring out ways of needing them. Not entirely sure what that means.

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,879 @@
...
+
+            Ok(TypedRange {
+                datatype,
+                range: Range::from((start, end)),
+            })
+        })
+    }
+}
+
+pub struct FragmentInfo<'ctx, 'info> {
```
rroelke, line 657: Might as well derive ContextBound here

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,879 @@
...
+        })?;
+
+        if dense == 1 {
+            Ok(FragmentType::Dense)
+        } else {
+            Ok(FragmentType::Sparse)
+        }
+    }
+
+    pub fn timestamp_range(&self, frag_idx: u32) -> TileDBResult<(u64, u64)> {
```
rroelke, line 249: For the return type - for things like extents we have done `[T; 2]`. THis would be another item to make consistent.
At least as far as JSON is concerned the slice approach gives you something a little nicer, since you would have
`{ "timestamp_range": [start, end] }`
instead of
`{ "timestamp_range": { "0": start, "1": end }}` (which is what I assume the tuple type turns into)

The slice approach also lets you do things like iterate but I'm not really sure why one would want to do that

tiledb/api/src/array/fragment_info.rs
```diff
@@ -0,0 +1,879 @@
...
+    num_fragments: u32,
+    index: u32,
+}
+
+impl<'ctx, 'info: 'ctx> Iterator for FragmentInfoListIterator<'ctx, 'info> {
+    type Item = FragmentInfo<'ctx, 'info>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index >= self.num_fragments {
+            return None;
```
rroelke, line 801: Looks like you can `impl FusedIterator`
https://doc.rust-lang.org/std/iter/trait.FusedIterator.html

rroelke (approved):
Made another pass, +1 !
I made a few minor suggestions but nothing that would merit another review.

tiledb/api/src/error.rs
```diff
@@ -67,6 +67,9 @@ pub enum Error {
     /// Error when a function has an invalid argument
     #[error("Invalid argument: {0}")]
     InvalidArgument(#[source] anyhow::Error),
+    /// Error when an invalid index is provided as an arguemnt.
+    #[error("Invalid index: {0}")]
+    InvalidIndex(usize),
```
rroelke, line 72: Looks unused? At least CTRL-F doesn't see anything

tiledb/api/src/array/fragment_info.rs
davisp: Good call. Fixed.

tiledb/api/src/error.rs
davisp: https://github.com/TileDB-Inc/tiledb-rs/blob/42b1d87823b0938bbb3454f8b000ab6bbb616d01/tiledb/api/src/array/fragment_info.rs#L631

tiledb/api/src/array/fragment_info.rs
davisp: Neat! Done.

PR #54, opened 2024-04-12
Enable #[derive(ContextBound)]

rroelke:
This pull request enables writing `#[derive(ContextBound)]`.  The implementation looks for a field with an attribute either `#[context]`, which is the context field to return directly, or a field with attribute `#[base(ContextBound)]` which indicates that the impl should be deferred to the base field.

There are a few `unimplemented!()`s, but what *is* implemented is enough to replace all the manual impls in our current code, as well as satisfy the proc macro unit test which simulates how we will want to use this in the near future for the query API adapters.

### DIFF
--- a/tiledb/api/src/array/attribute/mod.rs
+++ b/tiledb/api/src/array/attribute/mod.rs
@@ -272,7 +272,7 @@ impl<'c1, 'c2> PartialEq<Attribute<'c2>> for Attribute<'c1> {
 
 #[derive(ContextBound)]
 pub struct Builder<'ctx> {
-    #[ContextBound]
+    #[base(ContextBound)]
     attr: Attribute<'ctx>,
 }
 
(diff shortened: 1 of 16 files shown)


tiledb/api/src/array/attribute/mod.rs
```diff
@@ -274,16 +270,12 @@ impl<'c1, 'c2> PartialEq<Attribute<'c2>> for Attribute<'c1> {
     }
 }
 
+#[derive(ContextBound)]
 pub struct Builder<'ctx> {
+    #[base(ContextBound)]
```
davisp, line 275: Oooh, nice!

davisp (approved):
+1. That slight tweak to `base(ContextBound)` is subtle but effective in making that case more obvious.

PR #56, opened 2024-04-15
Fix domain shrinking proptest

### LINKED ISSUE #55: The Nightly Build Failure job failed on Friday (2024-04-12)
The Nightly Build Failure job failed on Friday (2024-04-12) in run [8670208978](https://github.com/TileDB-Inc/tiledb-rs/actions/runs/8670208978)

davisp:
There were two underlying but related bugs in the dimension range and extent generation. The first bug found and fixed was when we had floating point values with sufficiently large exponents that caused this relationship: `x + y == x`. I.e., "a vary large value + 1 == the same value". This is due to floating point resolution at resolution being larger than 1 when the exponent is large enough. For this bug, the end result was that our extent bounds ended up being `[1.0, 0.0)` which is an empty range that causes proptest to panic internally.

A second version of this same bug also existed for integral values when we had the case of `x + 1 = y`. In this case, due to the open rhs of a range, we ended up with `[1.0, 1.0)` which is also empty.

Once those were fixed, the next issue was based on the same underlying issue in that `x + y == x` can be true for floating point values when `y` is sufficiently smaller than `one`. For this case I had to implement the SmallestPositiveValue trait which is `1` for integral types as would be expected, but is `{f32,f64}::MIN_POSITIVE` which is a tiny value of approximately `1e-127`.

Also, as a side note, since this test is testing the actual proptest shrinking strategy, the TestRunner test cases value is ignored. The easiest way to trigger the issue was to just wrap the test in a `for _ in range 0..1000 { ... }` loop.

Fixes #55

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -431,6 +431,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "float_next_after"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8bf7cc16383c4b8d58b9905a8509f02926ce3058053c056376248d958c9df1e8"
+
+[[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


rroelke (approved):
Looks good, we may even need these new traits for other things in the future. Thanks!

PR #57, opened 2024-04-15
Add API Coverage CI Job

davisp:
This fixes all of our existing API discrepancies and requires that no more are introduced (i.e., CI will fail if whatever declarations in `tiledb/sys` don't exactly match what bindgen generates unless they are remapped).

I've also fixed up the output so we can use it as a build status on GitHub.

### DIFF
--- a/.github/workflows/nigthtly-ci.yml
+++ b/.github/workflows/nigthtly-ci.yml
@@ -1,10 +1,9 @@
+name: Nightlies
 on:
   schedule:
     # runs every day at 1:45 UTC
     - cron: "45 01 * * *"
   workflow_dispatch:
-
-name: Nightlies
 
 jobs:
   update_build_and_test:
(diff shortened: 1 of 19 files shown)


davisp:
An example summary can be found here: https://github.com/davisp/tiledb-rs/actions/runs/8696383958#summary-23849531109

davisp:
@ihnorton I'm just asking you to skim the GHA workflows to see if you notice anything glaringly wrong. You might also take a gander at the generated report as well in case you have any suggestions.

@rroelke If you don't care about reviewing, don't bother. But I did make some minor changes to tiledb/sys and tiledb/api as part of this PR to resolve all the differences between bindgen and what we declare in tiledb/sys. For the most part this just meant changing `u32` back into the appropriate `tiledb_datatype_t` or whatever the case may have been. Though VFSMode got the standardization treatment.

rroelke (approved):

PR #58, opened 2024-04-16
Move arrow sub-crate into api crate feature

rroelke:
We may to do some heavier work with arrow integrations, such as including implementing some of our traits for arrow structures. Rust forbids implementing a trait on a struct unless the trait is defined in the current crate, or the struct is defined in the current crate. Hence we should move our arrow work into the api crate and hide it behind a feature.

Of note here is that the arrow tests all used our proptest strategies.  Cargo does not allow features to specify their own dev dependencies, so instead we change all the strategy modules from
`#[cfg(feature = "proptest-strategies")]`
to
`#[cfg(any(test, feature = "proptest-strategies"))]`
so that the strategies can be used whether testing or specifying tiledb as a dependency with the proptest-strategies feature.


tiledb/api/src/array/attribute/mod.rs
```diff
@@ -520,7 +520,10 @@ impl<'ctx> Factory<'ctx> for AttributeData {
     }
 }
 
-#[cfg(feature = "proptest-strategies")]
+#[cfg(feature = "arrow")]
```
davisp, line 523: Should we not add the same `any(test, ...)` for arrow?

davisp (approved):
+1

tiledb/api/src/array/attribute/mod.rs
rroelke: I'd say the difference is that the proptest strategies we need to do the testing of the other stuff, whereas arrow is just a thing we want to test.

PR #59, opened 2024-04-16
Replace dev with main in CI

davisp:
I just noticed I copy pasta'ed the wrong default branch name when updating CI the other day.


davisp:
Fixup for [sc-45290]

PR #61, opened 2024-04-17
Fix subarray builder to correctly use C API

rroelke:
The previous builder was based on my misunderstanding of how to use the subarray API.  The pattern enforced by the builder prevented more than one dimension range from being used, because `SubarrayBuilder::dimension_range_typed` returned back the query builder.  Now it returns back the Subarray builder so that a more complex subarray can be constructed before changing back to the Query building state.

`cargo run --example quickstart_dense` before and after these changes shows that we now produce a correct result.

### DIFF
--- a/tiledb/api/src/query/subarray.rs
+++ b/tiledb/api/src/query/subarray.rs
@@ -3,7 +3,7 @@ use std::ops::Deref;
 use crate::array::DimensionKey;
 use crate::context::{CApiInterface, Context, ContextBound};
 use crate::convert::CAPIConverter;
-use crate::query::Builder as QueryBuilder;
+use crate::query::{Builder as QueryBuilder, Query};
 use crate::Result as TileDBResult;
 
 pub(crate) enum RawSubarray {
@@ -64,7 +64,7 @@ impl<'ctx> Builder<'ctx> {
         self,
         key: K,
         range: &[Conv; 2],
-    ) -> TileDBResult<QueryBuilder<'ctx>> {
+    ) -> TileDBResult<Self> {
         let c_context = self.subarray.context.capi();
         let c_subarray = *self.subarray.raw;
 
@@ -99,10 +99,11 @@ impl<'ctx> Builder<'ctx> {
                 })
             }
         }?;
-        self.build()
+        Ok(self)
     }
 
-    fn build(mut self) -> TileDBResult<QueryBuilder<'ctx>> {
+    /// Apply the subarray to the query, returning the query builder.
+    pub fn to_query(self) -> TileDBResult<QueryBuilder<'ctx>> {
         let c_context = self.subarray.context.capi();
         let c_query = *self.query.query.raw;
         let c_subarray = *self.subarray.raw;
@@ -110,7 +111,11 @@ impl<'ctx> Builder<'ctx> {
         self.capi_return(unsafe {
             ffi::tiledb_query_set_subarray_t(c_context, c_query, c_subarray)
         })?;
-        self.query.query.subarrays.push(self.subarray);
         Ok(self.query)
+    }
+
+    /// Apply the subarray to the query and finish constructing the query.
+    pub fn build(self) -> TileDBResult<Query<'ctx>> {
+        Ok(self.to_query()?.build())
     }
 }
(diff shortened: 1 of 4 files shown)


tiledb/api/src/query/subarray.rs
```diff
@@ -99,18 +99,23 @@ impl<'ctx> Builder<'ctx> {
...
 
         self.capi_return(unsafe {
             ffi::tiledb_query_set_subarray_t(c_context, c_query, c_subarray)
         })?;
-        self.query.query.subarrays.push(self.subarray);
         Ok(self.query)
     }
+
+    /// Apply the subarray to the query and finish constructing the query.
+    pub fn build(self) -> TileDBResult<Query<'ctx>> {
```
davisp, line 118: Wouldn't this force users to have to know that the subarray being set must go last on the query builder? That seems like a not very awesome requirement. I've stumbled on this "sub-builder" pattern a couple of times and generally end up avoiding it by using other means. Either by reshaping the "parent" builder's API or extracting things to accept a "pre-built" value.

I.e., I'd prefer something like either of these:

```rust

query.do_stuff()?
    .subarray(&[
        ("rows", &[1i32, 2]),
        ("columns" &[2i32, 4])
    ])
    .do_more_stuff()?;
    .build()?;
```

Or perhaps:

```rust
let subarray = SubarrayBuilder::new(&ctx)
    .dimension::<i32>("rows", 1, 2)?
    .dimension::<i32>("columns", 2, 4)?
    .build();

let query = builder.do_stuff()?
    .subarray(subarray)
    .do_other_stuff()?
    .build();
````

On the other hand, if the subarray forces our hand, I could also see having a different API for subarray as well.

```rust
let query = builder.do_stuff()?
    .start_subarray()
    .dimension::<i32>("rows", 1, 2)?
    .dimension::<i32>("columns", 2, 4))?
    .finish_subarray()?
    .do_other_stuff()?
    .build();

davisp (changes requested):
Assuming I've not read this wrong, I'd rather not have the SubarrayBuilder consuming the QueryBuilder. Forcing ordering on users seems like not great UX. And it also means that we can't ever have another "sub builder" type API since you wouldn't be able to use both.

tiledb/api/src/query/subarray.rs
rroelke: > Wouldn't this force users to have to know that the subarray being set must go last on the query builder?

This is not required as `to_query` can be used to get the query builder back.  I just made `build` do this as a convenience, but if it is a confusing convenience then I can revert it.  Where you suggest `start_subarray` and `finish_subarray`, what's in the PR is `add_subarray` and `to_query` - I like your nomenclature better.

davisp (approved):
+1

Just to check my understanding, the previous approach basically just had a `build()` method on SubarrayBuilder as convenience for when it was the last thing added to the query?

That actually seems like it might be a useful idea for the QueryCondition builder APIs since I imagine it'll be pretty easy to create something decently complex for building up boolean logic trees. Although here I prefer the start/finish pairing since its quite neatly expresses the push/pop of the builder stack.

rroelke:
> +1
> 
> Just to check my understanding, the previous approach basically just had a `build()` method on SubarrayBuilder as convenience for when it was the last thing added to the query?

Prior to the PR, on `main`, there was a similar sort of state machine with starting and finishing a subarray, but it was too aggressive because doing `dimension_range_typed` on the subarray gave you back the query builder state. So it was impossible to add multiple dimensions to the subarray.

The state of the branch as you initially reviewed it needed to add a new method which transitions the subarray builder back to the query builder.  I chose to name it `to_query`, and then thought it would be convenient just to put `build` in there to make the whole query so that if the subarray was the last thing you were doing then you could call `build()` instead of just `to_query().build()`.

> 
> That actually seems like it might be a useful idea for the QueryCondition builder APIs since I imagine it'll be pretty easy to create something decently complex for building up boolean logic trees. Although here I prefer the start/finish pairing since its quite neatly expresses the push/pop of the builder stack.

Probably a `finish_condition` which unwraps all the way back up to the query would be fine, no?  I don't see how the "convenience" would help this

davisp:
> Probably a `finish_condition` which unwraps all the way back up to the query would be fine, no? I don't see how the "convenience" would help this

Oh, I'm just thinking that it'll likely be quite possible that there'll get to be some deep nesting so being able to say "I'm done" at any stage and have it close everything up automatically could be nice. I'll see how it shakes out when I get to wrapping those APIs.

PR #62, opened 2024-04-17
Improve field lookup from schema API

rroelke:
First we use the `DimensionKey` (soon to be re-named, from the fragment info PR) to enable looking up an attribute by name.  Then we also add `crate::array::schema::Field` to encapsulate commonalities to Dimension and Attribute, for use cases where it does not matter whether a schema field is a dimension or attribute.


davisp (approved):
Nice!

PR #67, opened 2024-04-18
Wrap QueryCondition APIs

davisp:
This wraps the query condition APIs with a set of helpful builder utilities for creating query conditions from Rust. Notably, I have not added the `Query::set_condition` APIs yet as the Query type is undergoing some major surgery as part of PR #51. Once that is merged we can come back and add the simple `set_condition` implementations.


davisp:
@rroelke Heads up, I'd start this review by reading the tests to show how things are intended to be used.

tiledb/api/src/query/conditions.rs
rroelke, line 1: This doesn't strike me as the right way to approach this. I think we probably ought to have a cleaner separation of building an expression tree and invoking the C APIs.  This keeps most of the state on the C side, which as you already have brought up in discussion means we get no introspection.  That means no checking if what's on the C side is correct, but it also means that there's no pattern matching on the query condition, or debug printing, or any way to construct a `datafusion_expr::Expr` from a query condition (we will definitely go from `datafusion_expr::Expr` => query condition, I don't know whether we'll ever need to go the other direction, but this API forbids that).

Using an `enum` to represent a more traditional expression tree, probably something resembling `datafusion_expr::Expr`, might save a lot of the boilerplate here, but it would also solve those problems and make this a much more useful API.

```
enum QueryConditionExpr {
    Literal(ScalarValue),
    Variable(String),
    BinaryExpr(BinaryOperation, Box<QueryConditionExpr>, Box<QueryConditionExpr>),
    ...
}
enum QueryCondition {
    /* subset of the above with only the items that are allowed at the top level */
}

impl QueryBuilder {
    fn condition(self, condition: QueryCondition) -> TileDBResult<Self> {
        /* alloc the condition, walks the condition tree to do the C API calls, attaches to query, frees the condition */
    }
}
```

One of the decisions we have to continuously make as we write the bindings is how much state to duplicate on the Rust side vs. keep inside of core.  I think the main objection to the `enum` suggestion is that it duplicates all of the state.  But this implementation also does that - it just makes the API calls in between the construction instead of doing them all at once.


EDIT:
I wrote all of the above and then read it over again, and I think what you have with the `Initializer` structures is pretty similar to the enum idea, as it does build a tree structure and then fold it into the QueryCondition at the end.  But the use of `Box<dyn QueryConditionInitializer>` has several disadvantages over the enum approach as it prevents pattern matching and debug printing.  And I suppose you could add `Send` to the interior as well but right now it is not.

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,736 @@
...
+    fn initialize<'ctx>(
+        &self,
+        ctx: &'ctx Context,
+    ) -> TileDBResult<QueryCondition<'ctx>> {
+        let cond = self.initializer.initialize(ctx)?;
+        cond.negate()
+    }
+}
+
+pub struct QC {
```
rroelke, line 415: This is a variable right?  It should be named variable

rroelke (changes requested):

tiledb/api/src/query/conditions.rs
davisp: Totes agree that the lack of reading things back means we either get to store state in Rust (which as you note we've been trying to avoid).

The first thing I went for was the `Box<dyn QueryConditionInitializer>` because that just happened to be the first that came to mind. Skimming the datafusion_expr crate I can see removing the function pointers in favor of an enum variant tree with boxes. Which sounds about Rust.

I'll give it a whirl and post what I find after poking at it.

davisp:
Swapped to draft status to avoid accidentally merging this while it depends on the validity vector PR.

tiledb/api/src/query/conditions.rs
davisp: @rroelke I've gone and rewritten the conditions.rs module using an enum variant approach. I agree that it feels better than the nest of dynamic traits I went with on the first pass. Give it a read and let me know what you think. I'll wait to hear from you before I jump into adding a bunch of tests using your array data proptest strategies.

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,736 @@
...
+                c_val.as_ptr() as *const std::ffi::c_void,
+                std::mem::size_of_val(&c_val) as u64,
+                op.capi_enum(),
+            )
+        })?;
+
+        Ok(cond)
+    }
+
+    fn new_string<T: AsRef<str>>(
```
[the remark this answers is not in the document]
davisp: The enum variants actually cleaned this up quite nicely. Although now we're left with the decision on whether the new Value from Groups should be combined with this variant. I'm currently leaning towards "probably not?" with a side of "but I could see it happening".

tiledb/api/src/query/conditions.rs
davisp: I'm not sure what you mean by this. I assume the `new` method was a bit misleading though. A QC is meant to represent any query condition whether it be a simple `x < 5` comparison or some complicated nested tree of boolean algebra. The new implementation has a hopefully less confusing API that looks similar but with better types.

At the bottom of the type tree is a Condition which has the basic shape of `field operator value`. There are three variants to enforce correctness:

1. CmpCondition - field {<, <=, ==, !=, >=, >} value which is a single value
2. SetMemberCondition field {in, not_in} value which is a vec of values
3. NullCondition: field {is_null, not_null} (value elided cause its nullptr)

After conditions, we have the `QC` structure which has three variants:

1. Cond - One of the three condition types from above
2. Comb (lhs: QC, rhs: QC, operator) <- two sub conditions with a combination operator (AND or OR)
3. Negate (val: QC) <- Applies !QC

The final bit is the syntactical sugar for creating and combining query conditions. Creating a condition starts with `QC::field("foo")` which returns a `Field` struct. A `Field` just contains a copy of the field name in a `String` and provides a number of helper methods to generate any of the three variant types. Thus you can do any of the following:

```rust
let qc1 = QC::field("foo").lt(5);
let qc2 = QC::field("bar").in(&[1, 2, 3, 4][..]);
let qc3 = QC::field("baz").is_null();
```

Then for combining and negating, there are the `BitAnd`, `BitOr`, and `Not` which allow for things like:

```rust
let qc4 = qc1 & qc2;
let qc5 = qc2 | qc3;
let qc6 = !qc5;
```

When we get to predicate push down, we'll also probably add more programatic methods for conversion. Something like the following. I've not written these yet though because I don't want to add them before we need them as well as I could see slight variations depending on what the other side of the conversion process looks like.

```rust
let qc1 = QC::field("foo").cmp(CmpOp::Less, 5);
let qc2 = QC::field("bar").set_member(SetMemberOp::In, &[1, 2, 3, 4, 5][..]);
let qc3 = QC::field("baz").nullness(NullnessOp::IsNull);
let qc4 = qc1.combine(qc2, CombOp::And);
let qc5 = qc2.combine(qc3, CombOp::Or);
let qc6 = qc5.negate();
```

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+                self.op.capi_enum(),
+            )
+        })?;
+
+        Ok(cond)
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum Condition {
```
rroelke, line 452: "Predicate" maybe?

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+    pub fn not_null(self) -> QC {
+        QC::Cond(Condition::Null(NullnessCondition {
+            field: self.field,
+            op: NullnessOp::NotNull,
+        }))
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum QC {
```
rroelke, line 557: I really don't like "QC" as a name here, if the examples want to re-export a name as `QC` then that seems fine but I definitely feel like the proper type name should be more verbose.

Is there a reason you didn't move the `Condition` variants into this, and just have one enum?

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+                    context: ctx,
+                    raw: RawQueryCondition::Owned(c_cond),
+                })
+            }
+            Self::Negate(expr) => {
+                let mut c_cond: *mut ffi::tiledb_query_condition_t = out_ptr!();
+                ctx.capi_return(unsafe {
+                    ffi::tiledb_query_condition_negate(
+                        ctx.capi(),
+                        expr.build(ctx)?.capi(),
```
rroelke, line 605: `expr.build(ctx)` is not unsafe, I think you should pull this out of the `unsafe` block.
It also will help future readers who ask the question "wait, don't we have to build the `expr` first? Where is that? Oh wait there it is"

rroelke: As a generalization of this, I think most of our code kind of trains readers to ignore most of the FFI arguments because they are generally just simple `let` bindings.  I don't think we should deviate that

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn basic_op_test() -> TileDBResult<()> {
+        let qc1 = QC::field("field").lt(5);
+        let qc2 = QC::field("field").ge(6);
+        let qc = qc1 | qc2;
```
rroelke, line 690: IMO these tests should do a bit of basic pattern matching now, just to be sure that you actually got the right thing out.

rroelke (changes requested):

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+    value: SetMemberValue,
+}
+
+impl SetMemberCondition {
+    fn build<'ctx>(
+        &self,
+        ctx: &'ctx Context,
+    ) -> TileDBResult<QueryCondition<'ctx>> {
+        // First things first, sets require a non-zero length vector. I would
+        // prefer if we couldn't even create SetMemberValues with zero length
```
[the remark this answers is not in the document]
rroelke: I don't think it's necessary to use one here but maybe we can file a follow-up story.

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+cmp_value_impl!(f64, CmpValue::Float64);
+cmp_value_impl!(String, CmpValue::String);
+
+impl From<&str> for CmpValue {
+    fn from(val: &str) -> CmpValue {
+        CmpValue::String(val.to_string())
+    }
+}
+
+// N.B. I initially tried slices here, but that breaks the Deserialize trait.
```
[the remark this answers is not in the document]
davisp: Right, I initially tried using `&[T]` but that caused the Deserializer trait to complain that arbitrary length slices aren't supported. I assumed it was an ownership thing because during deserialization who would own the slices?

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum SetMemberValue {
```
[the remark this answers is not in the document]
davisp: > I assume what's meant here is that each element of the Vec is an element to test for the set membership.

Exactly correct.

> Is this complete, then?

Yep. Query conditions don't support arbitrary `cell_val_num > 1` attributes. The only var sized support is for StringAscii and StringUtf8.

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+            Self::UInt16(val) => slice_to_ptr_and_size!(val),
+            Self::UInt32(val) => slice_to_ptr_and_size!(val),
+            Self::UInt64(val) => slice_to_ptr_and_size!(val),
+            Self::Int8(val) => slice_to_ptr_and_size!(val),
+            Self::Int16(val) => slice_to_ptr_and_size!(val),
+            Self::Int32(val) => slice_to_ptr_and_size!(val),
+            Self::Int64(val) => slice_to_ptr_and_size!(val),
+            Self::Float32(val) => slice_to_ptr_and_size!(val),
+            Self::Float64(val) => slice_to_ptr_and_size!(val),
+            Self::String(_) => (std::ptr::null(), 0),
```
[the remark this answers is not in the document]
davisp: I don't think going down to `Vec<Vec<u8>>` is correct here based on the QueryCondition implementation constraints. Though I see your point and the Option is definitely cleaner and more Rust-y so I'll make that switch. In hindsight this seems like my brain pulling in C++'isms rather than updating to Rust Option result style.

tiledb/api/src/query/conditions.rs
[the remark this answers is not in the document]
davisp: Arguably, we could also just relax the constraints on the query condition side to allow testing for set membership against the empty set which in hindsight seems perfectly logical. I wrote the rejection of empty sets as more of a UX thing thinking that an empty set was more likely a programmer error.

tiledb/api/src/query/conditions.rs
davisp: That's fair. I'll write out the QueryCondition name for the implementation and then provide a re-exported `QC` alias.

tiledb/api/src/query/conditions.rs
davisp: That's fair. Will update.

tiledb/api/src/query/conditions.rs
davisp: Totes fair. Will update.

tiledb/api/src/query/conditions.rs
```diff
@@ -0,0 +1,754 @@
...
+            }
+            CombineOp::Or => {
+                ffi::tiledb_query_condition_combination_op_t_TILEDB_OR
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum CmpValue {
```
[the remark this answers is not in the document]
davisp: Literal seems reasonable. I'm also not opposed to renaming anything in here when we get to datafusion conversions if it helps clear things up.

tiledb/api/src/query/conditions.rs
davisp: Ooops. I went with `QueryConditionExpr` because of the name clash with the actual `QueryCondition`.

tiledb/api/src/query/conditions.rs
davisp: So, that point actually made me come around to pre-declaring all of the FFI arguments outside of unsafe (other than calling `as_ptr()` on anything related to `cstring!(foo)`). So I'll definitely keep doing that going forward.

Also, as I went to move that line, I suddenly realized I think that was actually a super subtle bug that we might want to think about checking things in CI with ASAN. I'll check if that's an easy thing to turn on. But basically, I'm pretty sure that was a "returning pointer to temporary that doesn't live long enough" but Rust can't tell. The `build(ctx)?` return value is a temporary there, and then immediately calling `capi()` on it means that it the actual underlying C++ query condition would get deleted before the pointer is passed to FFI. I'm not sure how drop works in unsafe blocks so I'm not sure if we were being saved by something specific there, or perhaps we just got lucky that C++ could read the bit of memory due to shenanigans.

Either way, the result of `build` is now a temporary let assignment. I just wonder if there's something we could do to tell the compiler that the results of capi calls has to live shorter than the object. Do you think pointers can have lifetimes? Looks like the closest we can get would be to return our `RawThing` structs and then deref them since it seems Rust won't do lifetimes on pointers: https://stackoverflow.com/a/27940991

rroelke (approved):

PR #69, opened 2024-04-19
Groups API  + Tests + Example v2

abigalekim:


tiledb/api/examples/groups.rs
```diff
@@ -0,0 +1,151 @@
...
+    )?;
+    create_array(
+        "my_group/sparse_arrays/array_D",
+        tiledb::array::ArrayType::Sparse,
+    )?;
+
+    // Add members to groups
+    let mut group = Group::open(&tdb, "my_group", tiledb::query::QueryType::Write)?;
+
+    //print_group()?;
```
rroelke, line 111: Remove before merging

tiledb/api/src/context.rs
```diff
@@ -1,16 +1,50 @@
...
+}
+
+impl Debug for ObjectType {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        <Self as Display>::fmt(self, f)
+    }
+}
+
+impl Display for ObjectType {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
```
rroelke, line 39: Porting this comment from the other PR - this should parallel what you did for `Mode`

tiledb/api/src/error.rs
```diff
@@ -84,6 +86,64 @@ impl Display for DatatypeErrorKind {
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum ObjectTypeErrorKind {
+    InvalidDiscriminant(u64),
+    TypeMismatch {
+        user_type: &'static str,
```
rroelke, line 93: In this case you could probably do `user_type: ObjectType`.
For datatype this field is something like `u32` or `f64` or what have you, so the mismatch is between the user programattic type and the TileDB data type, which isn't the same as the mismatch here

tiledb/api/src/error.rs
```diff
@@ -84,6 +86,64 @@ impl Display for DatatypeErrorKind {
...
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum ModeErrorKind {
+    InvalidDiscriminant(u64),
+    TypeMismatch {
+        user_type: &'static str,
```
rroelke, line 122: Same as above

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,157 @@
...
+    };
+}
+
+impl Value {
+    pub fn c_vec(&self) -> (*const std::ffi::c_void, usize) {
+        value_typed!(self, _DT, vec, get_value_vec(vec))
+    }
+}
+
+pub trait ValueType {
```
rroelke, line 80: This should be `std::convert::From`

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,157 @@
...
+
+#[derive(Debug)]
+pub struct Metadata {
+    pub key: String,
+    pub datatype: Datatype,
+    pub value: Value,
+}
+
+impl Metadata {
+    pub fn create<T>(key: String, datatype: Datatype, vec: Vec<T>) -> Self
```
rroelke, line 115: "create" sounds like some work is going on, this should probably just be called `new`

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,157 @@
...
+        T: ValueType,
+    {
+        return Metadata {
+            key,
+            datatype,
+            value: T::get_value(vec),
+        };
+    }
+
+    pub fn new(
```
rroelke, line 126: I don't think we should enable any external APIs with `ffi`.  So this should probably just be `pub(crate)`.

`from_raw` should probably also be part of the name, since that's what this is doing.

This function should be marked `unsafe` since it is not safe on all possible inputs.

But overall I'm not sure you need this - there's just one call site right?  You could just put the `fn_typed` there calling `Metadata::new`

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,157 @@
...
+        });
+
+        Metadata {
+            key,
+            datatype,
+            value,
+        }
+    }
+
+    pub fn c_data(
```
rroelke, line 150: Per my above comment I think this should be at most `pub(crate)`

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,559 @@
...
+        let c_context = self.context.capi();
+        let c_group = Self::capi(self);
+
+        let mut c_open: i32 = out_ptr!();
+        self.context
+            .capi_return(unsafe {
+                ffi::tiledb_group_is_open(c_context, c_group, &mut c_open)
+            })
+            .expect("TileDB internal error when checking for open group.");
+        if c_open > 0 {
```
rroelke, line 551: In my opinion it should not be possible to construct a `Group` if the group is *not* open.

I'm fine with this being present as a failsafe mechanism *somewhere* but:
1) I think it should probably  be in `open`, to prevent constructing the `Group` (unless there are APIs which are valid to call on a closed group? is that the case?); and/or
2) it should be remarked upon with a comment that we don't really expect it ever to find a closed group.

rroelke (changes requested):
Seems like a lot of the comments I left on the previous PR still apply here.

tiledb/api/src/group/mod.rs
abigalekim: Wait no, the Group is constructed in create, and then opened once it is created. And the function open itself returns a Self, so by construction it is open? And close is called upon dropping an object.

rroelke: I should be clearer - a user of the API should not be able to get a Group type object unless the group it represents is open.  Your code currently does this, since `Group::open` returns `Err` if `tiledb_group_open` fails.  So the check here to `tiledb_group_is_open` in the destructor is kind of misleading, because it communicates that we might have a Group type object which represents a non-open group.

Unless the group can spontaneously close - I sure hope it can't!

I think you can add clarity to the use of `tiledb_group_is_open` by either calling it in `open` to prevent `open` from returning successfully, or by leaving a hard-to-miss comment here that we don't ever expect the group to be closed at this point.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+        &mut self,
+        uri: S,
+        relative: bool,
+        name: Option<T>,
+    ) -> TileDBResult<()>
+    where
+        S: AsRef<str>,
+        T: AsRef<str>,
+    {
+        println!("{}", uri.as_ref());
```
rroelke, line 148: Remove any `println`s before merging

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+    pub fn is_open(&self) -> TileDBResult<bool> {
+        let c_context = self.context.capi();
+        let mut c_open: i32 = out_ptr!();
+        self.context.capi_return(unsafe {
+            ffi::tiledb_group_is_open(c_context, Self::capi(self), &mut c_open)
+        })?;
+        Ok(c_open > 0)
+    }
+
+    pub fn dump(&self, recursive: bool) -> TileDBResult<Option<String>> {
```
rroelke, line 282: Does this ever return `None`?

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+            raw: RawConfig::Owned(c_cfg),
+        })
+    }
+
+    pub fn put_metadata(&mut self, metadata: Metadata) -> TileDBResult<()> {
+        println!("{:?}", metadata);
+        let c_context = self.context.capi();
+        let (vec_size, vec_ptr, datatype) = metadata.c_data();
+        println!("{:?}", metadata);
+        let c_key = cstring!(metadata.key.clone()); // we're partially moving metadata
```
rroelke, line 330: Once you remove the `println`s I think you probably won't need the `clone` any more

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+        if exists == 0 {
+            return Ok(None);
+        }
+
+        let datatype = Datatype::try_from(c_datatype)?;
+        Ok(Some(datatype))
+    }
+
+    pub fn consolidate_metadata<S>(
+        &self,
```
rroelke, line 444: This C API doc doesn't mention anything about the group open mode but this sure sounds like a mutating function call...

Anyway, as I've been mulling over the whole "some methods are `&mut self` and some aren't" thing, it's kind of occurred to me that `&mut self` to represent a write mode method doesn't really work because the API allows you to just get another handle, and then you have ownership of that second handle and can call whatever you want anyway. And this can go in the direction of "just let the C API give back the error and don't worry about it" or in the direction of "embed the permissions to call certain APIs in the type system".  If we had infinite time I'd want the latter but let's move ahead with the former for now.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+        let mut group1_err =
+            Group::open(&tdb, group1_uri.to_owned(), QueryType::Read)?;
+        let res = group1_err.put_metadata(Metadata::new(
+            "key".to_owned(),
+            Datatype::Int32,
+            vec![5],
+        ));
+        assert!(res.is_err());
+
+        std::mem::drop(group1_err);
```
rroelke, line 547: I suppose it's something of a stylistic preference, but you can avoid this function call by using inner blocks instead.
```
{
    let group1_err = Group::open(&tdb, group1_uri.to_owned(), QueryType::Read)?;
    let res = group1_err.put_metadata(...);
    ...
    /* `res` will be dropped here, then `group1_err` */
}
```
I prefer this myself as it will drop everything used for this part of the test and be a better visual cue via indentation.

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Debug)]
+pub struct Metadata {
+    pub key: String,
+    pub datatype: Datatype,
+    pub value: Value,
+}
+
+impl Metadata {
+    pub fn new<T>(key: String, datatype: Datatype, vec: Vec<T>) -> Self
```
rroelke, line 110: I think this is probably the root cause of one of the bugs that we looked at, with the `f32` being a problem. This constructor doesn't require any correlation between `datatype` and `vec`.

I don't think we've really flushed out that story at all, but for now you can definitely catch *some* problems, like the one you encountered, by checking `std::mem::size_of::<T>()` against `datatype.size()` and making sure that they match.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+        let mut group_write =
+            Group::open(&tdb, group_uri.to_owned(), QueryType::Write)?;
+        group_write.delete_member("bb")?;
+        std::mem::drop(group_write);
+
+        let group_read =
+            Group::open(&tdb, group_uri.to_owned(), QueryType::Read)?;
+        let opt_string = group_read.dump(true)?;
+        match opt_string {
+            Some(s) => println!("{}", s),
```
rroelke, line 721: Is there an assert that can be used to express what this wants instead?
If it's hard to check the contents of the string then asserting none/some seems fine.

For the `dump` API to be useful anyway it probably need to be parsed to some useful representation. If it's not documented then all anyone can do with it is print it or check some/none as you are doing here.

tiledb/api/src/group/mod.rs
abigalekim: Yes, when the dump is empty.

tiledb/api/src/error.rs
abigalekim: Oh wait, yea sorry! Actually this invalid discriminant error is used in the C API conversion, since the CAPI stores this enum as ints under the hood.

tiledb/api/src/group/mod.rs
abigalekim: Fixed this!

tiledb/api/src/metadata.rs
abigalekim: Added this check, but keeping this comment since I think this may need to be an additional story/discussed.

tiledb/api/src/error.rs
rroelke: This comment is in reference to the `TypeMismatch` variant. I don't follow your comment.

Is `TypeMismatch` used anywhere? I checked out your branch and deleted it and it doesn't appear to be.  Better to remove it.  Looks the same for `ModeErrorKind`.

tiledb/api/src/group/mod.rs
rroelke: Oh, good point, you have a consuming method `delete_group` and then also this `Drop` impl.

Please leave a comment here though because it is not obvious why this is needed

tiledb/api/src/group/mod.rs
rroelke: Which LOC does that? I don't see anything returning `None`

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn new<T>(
+        key: String,
+        datatype: Datatype,
+        vec: Vec<T>,
+    ) -> TileDBResult<Self>
+    where
+        Value: From<Vec<T>>,
+    {
+        if std::mem::size_of::<T>() != datatype.size() as usize {
+            return Err(crate::error::Error::InvalidArgument(anyhow!(
```
rroelke, line 121: This should be `DataTypeErrorKind::TypeMismatch`

tiledb/api/src/metadata.rs
davisp: `datatype.is_compatible_type::<T>() -> bool` is probably what you want here.

tiledb/api/src/group/mod.rs
abigalekim: Fixed.

tiledb/api/src/group/mod.rs
abigalekim: Wrote a comment!

tiledb/api/src/error.rs
abigalekim: Deleted the type mismatch errors.

tiledb/api/src/metadata.rs
```diff
@@ -0,0 +1,167 @@
...
+    UInt64Value(Vec<u64>),
+    Int8Value(Vec<i8>),
+    Int16Value(Vec<i16>),
+    Int32Value(Vec<i32>),
+    Int64Value(Vec<i64>),
+    Float32Value(Vec<f32>),
+    Float64Value(Vec<f64>),
+    //StringAsciiValue(CString),
+    //BooleanValue(bool),
+    // maybe blobs?
```
davisp, line 21: Lets remove commented out code here.

davisp (commented):
Skimming this, I'm not seeing anything of concern. Just the one minor nit. I'll defer to @rroelke to approve since he's reviewed this more than me.

tiledb/api/src/group/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+        context.capi_return(unsafe {
+            ffi::tiledb_group_open(c_context, group_raw, query_type_raw)
+        })?;
+
+        let mut c_open: i32 = out_ptr!();
+        context
+            .capi_return(unsafe {
+                ffi::tiledb_group_is_open(c_context, raw_group.ffi, &mut c_open)
+            })
+            .expect("TileDB internal error when checking for open group.");
```
rroelke, line 107: `expect` isn't appropriate here since we can return `Result`

also I think you can remove this if you want, your comment yesterday (earlier this week? something) in `Drop` convinced me that you did it correctly the first time

rroelke (approved):
All righty!
Looks like there are two final comments but they should be pretty trivial to address.
Let's get this in there, nice work!

PR #70, opened 2024-04-20
Fix slow property tests

davisp:
There were two tests that were each taking over twenty seconds to complete. After some investigation with Ryan the other day, we narrowed it own to constantly reparsing the pattern passed to `string_regex`. The idea then was to put all of our `string_regex` patterns behind a lazy_static. However, it turns out that its none of our strategies causing the issue.

The issue is `vec(any::<String>(), 0..1024)`. That will constantly re-parse the default `string_regex` pattern of `\\PC*`. To avoid that slowness I've created a new strategy that generates `Vec<Vec<u8>>` and then prop_maps the inner vector into a String. This works because I'm only generating `u8` values in the range [1, 127) which is guaranteed to be 7-bit clean and infallibly convertible to UTF-8.

This knocks both of the twenty second tests to about three seconds.

### DIFF
--- a/tiledb/api/src/query/buffer.rs
+++ b/tiledb/api/src/query/buffer.rs
@@ -127,3 +127,21 @@ impl<'data, T> QueryBuffersMut<'data, T> {
         }
     }
 }
+
+#[cfg(any(test, feature = "proptest-strategies"))]
+pub mod strategy {
+    use proptest::collection::vec;
+    use proptest::prelude::*;
+
+    pub fn prop_string_vec(
+        range: proptest::collection::SizeRange,
+    ) -> impl Strategy<Value = Vec<String>> {
+        vec(vec(1u8..127, 0..64), range)
+            .prop_map(move |mut v| {
+                v.iter_mut()
+                    .map(|s| String::from_utf8(s.clone()).unwrap())
+                    .collect::<Vec<_>>()
+            })
+            .boxed()
+    }
+}
--- a/tiledb/api/src/query/read/callback.rs
+++ b/tiledb/api/src/query/read/callback.rs
@@ -653,7 +653,13 @@ mod tests {
 
     proptest! {
         #[test]
-        fn read_result_strings(record_capacity in MIN_RECORDS..=MAX_RECORDS, byte_capacity in MIN_BYTE_CAPACITY..=MAX_BYTE_CAPACITY, stringsrc in vec(any::<String>(), MIN_RECORDS..=MAX_RECORDS))
+        fn read_result_strings(
+            record_capacity in MIN_RECORDS..=MAX_RECORDS,
+            byte_capacity in MIN_BYTE_CAPACITY..=MAX_BYTE_CAPACITY,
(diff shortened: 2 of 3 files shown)


rroelke (approved):
A very nice improvement!

Approving, but I think in the future we should consider some means of getting multibyte characters in there, there can definitely be bugs associated with those which we would want to find

rroelke:
In the `proptest` source:

```
impl Arbitrary for String {
    type Parameters = StringParam;
    type Strategy = &'static str;

    /// ## Panics
    ///
    /// This implementation panics if the input is not a valid regex proptest
    /// can handle.
    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
        args.into()
    }
}
```

```
impl Strategy for str {
    type Tree = RegexGeneratorValueTree<String>;
    type Value = String;

    fn new_tree(&self, runner: &mut TestRunner) -> NewTree<Self> {
        string_regex(self).unwrap().new_tree(runner)
    }
}
```

Pretty easy to see that it will parse the regex again every time you want a new value.

This looks like it's probably not too challenging to patch? It certainly seems like a contribution worth making

davisp:
> Approving, but I think in the future we should consider some means of getting multibyte characters in there, there can definitely be bugs associated with those which we would want to find

I contemplated this but decided that given these tests aren't worried about the interpretation of the strings that showing roundtrip equality was enough here. Though I'd certainly agree that these aren't general replacements for the existing `\\PC*` approach.

> This looks like it's probably not too challenging to patch? It certainly seems like a contribution worth making

Well, you caught me off guard with that nerd snipe. When I started writing this comment I was thinking, theoretically, it wouldn't be impossible to implement `\\PC*` directly rather than going through the entire regex crate. However, that means I had to figure out precisely what `\\PC*` meant which proved challenging, yet interesting enough to full track down. And so here it is: `vec(one_of(char_class_list), 0..=32)` where `one_of(char_class_list)` is hand wavy pseudo code for [any code point in this list of ranges](https://gist.github.com/davisp/df6767abe5a99575364a27cd4e0896c4).

So not quite as simple as my original idea for `vec((0x20..0x10FFFF).prop_map(to_utf8 ....), 0..=32)`.

And I can unfortunately report that it doesn't look like the regex-syntax crate exports any of the required unicode knowledge in such a way that we can create those classes directly instead of using the parse/translate APIs it already provides. And beyond that, all of my attempts to optimize the use of the regex-syntax APIs fails to put much of a dent in the run time vs just parsing as is currently done.

Moral of the story, is that hand coding a static AST that is passed to the HIR translator directly only reduces runtime from 7.1s for a full parse to 6.8s for debug builds and an astounding 283ms vs ~240ms for release. I haven't looked into whether or not there's an opportunity to improve the regex-syntax internals beyond trying to just borrow its unicode tables.

That's enough unicode for one day.

rroelke:
Huh, I assumed that the construction of the regex strategy was taking lots of the time, but that does make more sense that it's actually the traversal of the data structures inside the strategy. Sorry for nerd snipe but thanks for closing that gap.

PR #71, opened 2024-04-23
Extend API to enable proptests which write arbitrary array data, and get started writing the proptests

rroelke:
This pull request takes the first serious look at using our query API to do something beyond the simple examples.  Generating an arbitrary write query means poking at an arbitrary schema's attributes, generating data to write into those attributes, generating a write query for them, and then also generating a read query to check that what we wrote is actually correct.

This pull request modifies various components so that there is a path to implementing all of the above, and then implements a proptest which generates a series of write queries and then does an array read at each point in the series to check that everything  was written correctly.

The read query API is heavily modified, as it previously had nigh-nonexistent support for dynamic queries.  We have that now with `VarRawReadQuery` and `CallbackVarArgReadQuery`, along with some other improvements.

### DIFF
--- a/tiledb/api/src/query/mod.rs
+++ b/tiledb/api/src/query/mod.rs
@@ -130,7 +130,10 @@ pub trait QueryBuilder<'ctx>: Sized {
 
     fn base(&self) -> &BuilderBase<'ctx>;
 
-    fn layout(self, layout: QueryLayout) -> TileDBResult<Self> {
+    fn layout(self, layout: QueryLayout) -> TileDBResult<Self>
+    where
+        Self: Sized,
+    {
         let c_context = self.base().context().capi();
         let c_query = **self.base().cquery();
         let c_layout = layout.capi_enum();
@@ -140,7 +143,10 @@ pub trait QueryBuilder<'ctx>: Sized {
         Ok(self)
     }
 
-    fn start_subarray(self) -> TileDBResult<SubarrayBuilder<'ctx, Self>> {
+    fn start_subarray(self) -> TileDBResult<SubarrayBuilder<'ctx, Self>>
+    where
+        Self: Sized,
+    {
         SubarrayBuilder::for_query(self)
     }
 
--- a/tiledb/api/src/query/mod.rs
+++ b/tiledb/api/src/query/mod.rs
@@ -175,16 +175,12 @@ impl<'ctx> QueryBuilder<'ctx> for BuilderBase<'ctx> {
 }
 
 impl<'ctx> BuilderBase<'ctx> {
-    fn new(
-        context: &'ctx Context,
-        array: Array<'ctx>,
-        query_type: QueryType,
-    ) -> TileDBResult<Self> {
-        let c_context = context.capi();
+    fn new(array: Array<'ctx>, query_type: QueryType) -> TileDBResult<Self> {
+        let c_context = array.context().capi();
         let c_array = **array.capi();
         let c_query_type = query_type.capi_enum();
         let mut c_query: *mut ffi::tiledb_query_t = out_ptr!();
-        context.capi_return(unsafe {
+        array.capi_return(unsafe {
             ffi::tiledb_query_alloc(
                 c_context,
                 c_array,
--- a/tiledb/api/src/query/mod.rs
+++ b/tiledb/api/src/query/mod.rs
@@ -74,6 +74,10 @@ impl<'ctx> QueryBase<'ctx> {
         })
         .map(|_| c_status)
     }
+
+    pub fn array(&self) -> &Array<'ctx> {
+        &self.array
+    }
 }
 
 impl<'ctx> Query<'ctx> for QueryBase<'ctx> {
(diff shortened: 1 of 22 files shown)


tiledb/api/src/query/mod.rs
```diff
@@ -74,12 +78,30 @@ impl<'ctx> QueryBase<'ctx> {
         })
         .map(|_| c_status)
     }
+
+    pub fn array(&self) -> &Array<'ctx> {
+        /*
+         * TODO: if this is pub then it defeats the purpose of the query API consuming
```
rroelke, line 84: Probably delete this comment, the main reason to consume the array is to prevent multiple queries running simultaneously on the same open array, and this `pub fn` will not enable that.

tiledb/api/src/query/read/callback.rs
```diff
@@ -497,6 +526,13 @@ macro_rules! query_read_callback {
                     }
                 }
             }
+
+            impl<'ctx, 'data, T, B> ReadQueryBuilder<'ctx, 'data> for $Builder<'data, T, B>
```
rroelke, line 530: (This is inside the macro for defining the static arg read callback stuff)

TODO: use the `VarRawReadQuery` as a basis for those, as that will repeat less code.

tiledb/api/src/query/read/callback.rs
```diff
@@ -535,6 +571,124 @@ query_read_callback!(
...
+                    .zip(buffers.iter())
+                    .map(|(&(nvalues, nbytes), buffers)| TypedRawReadOutput {
+                        nvalues,
+                        nbytes,
+                        buffers: buffers.as_shared(),
+                    })
+                    .collect::<Vec<TypedRawReadOutput>>();
+
+                let ir = callback_final.final_result(&args).map_err(|e| {
+                    let fields = vec![]; /* TODO */
```
rroelke, line 650: This shouldn't be a difficult TODO, just iterate over the `base` read handles.

tiledb/api/src/query/read/managed.rs
```diff
@@ -1,39 +1,164 @@
 use super::output::ScratchSpace;
 use super::*;
 
+pub enum ManagedScratch<'data> {
```
rroelke, line 4: This, and lots of other things, can be deleted by removing `register_constructor_managed` in favor of `ScratchStrategy`

tiledb/api/src/query/read/mod.rs
```diff
@@ -194,21 +194,33 @@ macro_rules! fn_register_callback {
...
                 T: $Callback
             {
                 let base = self;
                 $(
-                    let [< arg_ $U:snake >] = RawReadHandle::new(
-                        [< field_ $U:snake >], [< scratch_ $U:snake >]);
+                    let [< arg_ $U:snake >] = {
+                        let field = [< field_ $U:snake >];
+                        match [< scratch_ $U:snake >] {
+                            ScratchStrategy::AttributeDefault => unimplemented!(),
```
rroelke, line 214: This shouldn't be that hard to implement - get the array ref, get the schema ref, look up the attributes by name, call the "query allocator" function.

tiledb/api/src/query/write/strategy.rs
```diff
@@ -0,0 +1,979 @@
...
+            let ndimensions = self.schema.domain.dimension.len();
+            let nattributes = self.schema.attributes.len();
+
+            let dimensions_mask = match self.schema.array_type {
+                ArrayType::Dense => {
+                    /* dense array coordinates are handled by a subarray */
+                    vec![WriteFieldMask::Exclude; ndimensions]
+                }
+                ArrayType::Sparse => {
+                    /* sparse array must write coordinates */
```
rroelke, line 719: The strategy has to go further than this and restrict the range of what is written to the sparse dimensions. That's a big TODO in order to get this to pass

tiledb/api/src/query/mod.rs
davisp: Took me a second, but I'm pretty sure you mean that since its returning `&Array<'ctx>`, we can't create a query from that hence no worries about this being pub since it allows read access to the array without exposing the non-Sync'ness of our underlying Array type?

davisp (approved):
+1

Overall, everything looks good to me. There's obviously a bit of complexity here, but for the most part its in the proptest which we're shelving until later so I'm definitely not going to hold things up in review while I work through reading that code enough times to understand it top to bottom.

Bottom line, the code outside proptest all seems straightforward. Its obviously not creating huge changes to the outside APIs given the minimal changes to the examples. And the proptest demonstrates that the programatic interface is working. So a solid +1 from me.

tiledb/api/src/query/mod.rs
rroelke: Right, that is what I mean.

tiledb/api/src/query/write/strategy.rs
```diff
@@ -0,0 +1,948 @@
...
+                    );
+                }
+            }
+        }
+    }
+}
+
+/// Mask for whether a field should be included in a write query.
+// As of this writing, core does not support default values being filled in,
+// so this construct is not terribly useful. But someday that may change
```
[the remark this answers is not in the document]
rroelke: Interesting...yeah I definitely imagine we will want to exercise that

PR #72, opened 2024-04-23
Add subarray range API wrappers

davisp:
This is a draft PR to show the current approach to how I'm handling the range APIs. I'm gonna ask for a quick PR review to make sure that I'm not terribly off base.

I did find a few new macro patterns in @rroelke's latest PR that could reduce some of the repetitiveness in range.rs that I'll think harder about doing tomorrow with fresh eyes.


davisp:
@rroelke Can you give this a quick skim to see if anything jumps out at you?

Your `Foo: Type` macro thing is something I'm going to go back and apply all over in range.rs. I'm also contemplating replacing my `Range::as_u8_slices` with your enum variant macro approach that is similar to `fn_typed!`.

davisp:
@rroelke This is ready for review. I've managed to get fairly decent test coverage on the range.rs module last night.

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -223,6 +222,35 @@ impl<'ctx> Builder<'ctx> {
         })
     }
 
+    pub fn new_string(
```
rroelke, line 225: Ah, the one from the discussion yesterday.
A dimension has a domain and extent if and only if it is a dimension of integral or floating point type for a dense array... right?

And there's no indication in the API here that it's a dense array. Maybe there should be, maybe not - definitely would be a bit of a pain to re-do a bunch of the types to enforce these sorts of constraints. Might improve the code quality of the proptests.

Anyway I'm definitely not suggesting you do that, but if I got the "if and only if" right, then `new_string` is not a good name for this.  If it's optional then we should probably just pull it out of `Builder::new`, except we're beholden to calling `tiledb_dimension_alloc` with the domain/extent already known, so probably that means just using an Option arg on `new`?  `Option<DomainAndExtent>` or something, which has variants for each type so that there's no generic arg for the `new` function.

rroelke: This possible implementation of the `DomainAndExtent` is maybe a little annoying but not too pathological:
```
pub enum DimensionType {
    Int8(domain, extent),
    Int16(domain, extent),
    ...
    StringAscii
}

impl DimensionDatatype {
    pub fn datatype(&self) -> Datatype;
}

impl TryFrom<Datatype> for DimensionDatatype {
    ...
}
```
If we ever add variants to `Datatype` then they would be caught by a `match` clause in one of these functions, one assumes

tiledb/api/src/array/schema/mod.rs
```diff
@@ -21,13 +21,22 @@ use crate::Datatype;
...
+    OptionSubset,
+    PartialEq,
+    Serialize,
 )]
 #[cfg_attr(
     any(test, feature = "proptest-strategies"),
     derive(proptest_derive::Arbitrary)
 )]
 pub enum ArrayType {
+    #[default]
```
rroelke, line 39: I don't love this, as you probably know...
so the goal is really that we want to be able to do
```
SchemaData {
    /* other fields */
    ..Default::default()
}
```
i.e. do the initializer list for the fields we want to supply values for, with `Default::default()` to get the ones we don't care about.  And this piece is the chosen way to implement that.

This is kind of what the Builder pattern is for. `Builder::new` requires supplying the non-default args, and then everything else is a setter that returns `self`, so you end up with something that looks like of like the literal constructor.

tiledb/api/src/datatype/mod.rs
```diff
@@ -381,6 +386,56 @@ impl Datatype {
             || self.is_datetime_type()
             || self.is_time_type()
     }
+
+    #[cfg(test)]
```
rroelke, line 390: Huh, I don't think I've seen this except for a `mod tests` item before, but this makes sense. Neat.

tiledb/api/src/datatype/mod.rs
```diff
@@ -529,40 +584,40 @@ macro_rules! fn_typed {
...
             Datatype::Float32 => { type $typename = f32; $then },
             Datatype::Float64 => { type $typename = f64; $then },
-            Datatype::Char => unimplemented!(),
-            Datatype::StringAscii => unimplemented!(),
-            Datatype::StringUtf8 => unimplemented!(),
-            Datatype::StringUtf16 => unimplemented!(),
-            Datatype::StringUtf32 => unimplemented!(),
-            Datatype::StringUcs2 => unimplemented!(),
-            Datatype::StringUcs4 => unimplemented!(),
+            Datatype::Char => { type $typename = i8; $then },
```
rroelke, line 587: Seems fine for now but I sorta wonder if in the future we should use newtypes for these so that we can do custom display, and etc.

And also now that I'm playing more with the Arrow datatypes I wonder if we should draw some inspiration from them too.

Just musing. Looks good for now

tiledb/api/src/query/subarray.rs
```diff
@@ -33,6 +39,121 @@ pub struct Subarray<'ctx> {
...
+    }
+
+    pub(crate) fn new(context: &'ctx Context, raw: RawSubarray) -> Self {
+        Subarray { context, raw }
+    }
+
+    /// Return all dimension ranges set on the query.
+    pub fn ranges(
+        &self,
+        schema: &'ctx Schema,
```
rroelke, line 54: Can a subarray exist without a query?  Maybe instead of an `&'ctx Context` member subarray should have a `&'query Query` member.  THen you can fish the array out of the query and fish the schema out of the array.

tiledb/api/src/query/subarray.rs
```diff
@@ -33,6 +39,121 @@ pub struct Subarray<'ctx> {
...
+                    let dtype = dim.datatype()?;
+                    let cvn = dim.cell_val_num()?;
+                    let range =
+                        TypedRange::from_slices(dtype, cvn, &start, &end)?
+                            .range;
+                    dim_ranges.push(range);
+                } else {
+                    let dtype = dim.datatype()?;
+                    let cvn = dim.cell_val_num()?;
+                    let size = match cvn {
```
rroelke, line 117: `u32::from(cvn)` is also possible and a bit more concise

rroelke: Alternatively you could do an outer `match` instead of `dim.is_var_sized()`

rroelke (commented):
Will come back to this Monday morning...

tiledb/api/examples/multi_range_subarray.rs
```diff
@@ -0,0 +1,100 @@
...
+        .add_range("rows", &[1, 2])?
+        .add_range("rows", &[4, 4])?
+        .add_range("cols", &[1, 4])?
+        .finish_subarray()?
+        .build();
+
+    let (a, (cols, (rows, ()))) = query.execute()?;
+
+    for (row, col, a) in izip!(rows, cols, a) {
+        println!("{} {} {}", row, col, a);
```
rroelke, line 42: What is the expected result supposed to be? Can you add comments?

tiledb/api/src/query/subarray.rs
```diff
@@ -33,6 +39,121 @@ pub struct Subarray<'ctx> {
...
+                        TypedRange::from_slices(dtype, cvn, &start, &end)?
+                            .range;
+                    dim_ranges.push(range);
+                }
+            }
+
+            ranges.push(dim_ranges);
+        }
+
+        Ok(Vec::new())
```
rroelke, line 153: `Ok(ranges)`, no?
Did this not get caught in tests?

tiledb/api/src/query/subarray.rs
```diff
@@ -63,45 +184,134 @@ where
...
+
+    /// Add a list of point ranges to the query.
+    pub fn add_point_ranges<Key: Into<LookupKey>, T: CAPISameRepr>(
+        self,
+        key: Key,
+        points: &[T],
+    ) -> TileDBResult<Self> {
+        let schema = self.query.base().query.array.schema()?;
+        let dim_idx = schema.domain()?.dimension_index(key)?;
+
```
rroelke, line 294: Some kind of datatype compatibility check oughta go here

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+                type $DT = f64;
+                $then
+            }
+        }
+    };
+}
+
+#[derive(Clone, Deserialize, Serialize, PartialEq)]
+pub enum MultiValueRange {
+    UInt8(Box<[u8]>, Box<[u8]>),
```
rroelke, line 131: I recall you prefer `Box` over `Vec` to avoid the excess capacity.  But a convenient way to create this sort of thing is by creating a `Vec` and then calling `into_boxed_slice`.

I was curious if an additional `truncate` would be required, but the reference doc for `into_boxed_slice` says it already does that - great!

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+}
+
+macro_rules! multi_value_range_try_from {
+    ($($V:ident : $U:ty),+) => {
+        $(
+            impl TryFrom<(u32, Box<[$U]>, Box<[$U]>)> for MultiValueRange {
+                type Error = crate::error::Error;
+                fn try_from(value: (u32, Box<[$U]>, Box<[$U]>)) ->
+                        TileDBResult<MultiValueRange> {
+                    if value.0 < 2 || value.0 == u32::MAX {
```
rroelke, line 157: Checking that I follow - `value.0` is meant to be the length?  Is the idea with `value.0 < 2` that you should (must) use a `SingleValueRange` instead?

How come you don't just check `value.1.len()` and `value.2.len()` against 2 and against each other? That sounds more user-friendly.  Is it because the `u32` here holds a fixed cell val num?

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+            }
+        )+
+    }
+}
+
+var_value_range_from!(UInt8: u8, UInt16: u16, UInt32: u32, UInt64: u64);
+var_value_range_from!(Int8: i8, Int16: i16, Int32: i32, Int64: i64);
+var_value_range_from!(Float32: f32, Float64: f64);
+
+impl From<(&str, &str)> for VarValueRange {
```
rroelke, line 273: FWIW it's not totally useless to have an additional `From<(String, String)>`, as that could avoid a copy by using `String::into_bytes`.

Obviously this is not a performance bottleneck but figured I'd mention.

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+#[derive(Clone, Deserialize, Serialize, PartialEq)]
+pub enum Range {
+    Single(SingleValueRange),
+    Multi(MultiValueRange),
+    Var(VarValueRange),
+}
+
+impl Range {
+    // N.B. This is not `check_field_compatibility` because dimensions have
+    // restrictions on their cell_val_num that don't apply to attributes.
```
rroelke, line 346: We also don't apply subarray ranges to attributes... right?

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+    // See: tiledb::sm::Dimension::set_cell_val_num
+    pub fn check_dimension_compatibility(
+        &self,
+        datatype: Datatype,
+        cell_val_num: CellValNum,
+    ) -> TileDBResult<()> {
+        match self {
+            Self::Single(svr) => svr.check_datatype(datatype)?,
+            Self::Multi(_) => {
+                return Err(Error::InvalidArgument(anyhow!(
```
rroelke, line 357: This surprises me a bit, it seems like it would be just easier to support than Var. Oh well.

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+                type Error = crate::error::Error;
+                fn try_from(value: (u32, Box<[$U]>, Box<[$U]>)) -> TileDBResult<Range> {
+                    Ok(Range::Multi(MultiValueRange::try_from(value)?))
+                }
+            }
+        )+
+    }
+}
+
+range_multi_try_from!(UInt8: u8, UInt16: u16, UInt32: u32, UInt64: u64);
```
rroelke, line 437: These are all the same, you could consolidate by putting multiple `impl`s in the same macro expansion.

rroelke (changes requested):

tiledb/api/src/array/dimension/mod.rs
davisp: The if-and-only-if applies to the domain. There are also cases where extent is optional on non-string types, if memory serves I believe that's also sparse arrays so they can support floating point dimensions? Or perhaps any sparse dimension doesn't require an extent? Not sure on the particulars there. But easy enough to use your pattern there and just specify `Option<extent>`

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -329,36 +359,54 @@ impl<'ctx> Factory<'ctx> for DimensionData {
...
-                self.datatype,
-                &[d0, d1],
-                &extent,
-            )
-        })?;
+        let mut b = if self.datatype == Datatype::StringAscii {
+            Builder::new_string(context, &self.name, self.datatype)?
+        } else {
+            if self.domain.is_none() {
+                return Err(Error::InvalidArgument(anyhow!(
```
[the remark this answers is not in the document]
davisp: I'm not 100% actually. Though looking back, this is all wrong. The way to do this is to remove all of the logic here and go back to relying on core to tell us what's right and wrong. I'll rewrite this so that I can create a builder from a RawDimension and then add the FFI calls to just set what the user presented in the data. That coupled with your idea for the DimensionType enum means that most of the time it'll not get messed up on our end.

tiledb/api/src/array/schema/mod.rs
davisp: What do you think of adding `Invalid` as a variant that blows up as soon as we try and convert it? For datatypes I think we decided that the same behavior by using `Any` as the default would be acceptable? Granted this adds a new variant that has no meaning in C++ land, though, I'm not sure that concerns me.

tiledb/api/src/datatype/mod.rs
davisp: I don't remember what crate I was looking through ages ago, but they were using a similar pattern for cross platform code that was so awesome that it just stuck.

```rust
#[cfg(unix)]
fn do_thing(&self) {
    ....
}

#[cfg(windows)]
fn do_thing(&self) {
   ...
}
```

Having just spent a year working on a cross platform C++ library for a living, the simplicity of that made it stick pretty thoroughly.

tiledb/api/src/datatype/mod.rs
davisp: Yeah. In an ideal world our type system would get split between concrete and logical types though that's apparently contentious.

tiledb/api/src/query/subarray.rs
davisp: Huh! Good idea! The [`tiledb_subarray_alloc` requires an `tiledb_array_t*`](https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/c_api/tiledb.cc#L1739-L1742) so shoving that into the struct instead seems like a good approach.

tiledb/api/src/query/subarray.rs
davisp: Oh dear. How have I missed that convenience.... I'll got the from route.

tiledb/api/src/query/subarray.rs
davisp: Whoops. I hadn't added tests for this because I was waiting for the array strategy so that I'd just extend that. Then we shelved that for the time being. I'll throw in some local module tests to exercise this code so that I don't look so silly.

tiledb/api/src/query/subarray.rs
davisp: Good call. And easy enough.

tiledb/api/src/range.rs
davisp: `Box` over `Vec` isn't because of excess capacity (though it doesn't hurt), its because if we grab pointers to slices derived from Vec slices and pass them across the ffi boundary, they'll be invalidated as soon as the `Vec` reallocates which scared the dickens out of me. Using boxed slices means that as long as we have a ref to it, we're guaranteed that the underlying pointer is valid.

tiledb/api/src/range.rs
davisp: This would probably be more clear if that were correctly typed as a CellValNum. The logic here is saying "If you have a CellValNum::Fixed(1) or CellValNum::Var, go use the correct range type.

Fun fact, this isn't even used for domains because core doesn't support them, but it felt reasonable to include since it'd be kinda weird to not be able to create them at all.

tiledb/api/src/range.rs
davisp: I'll add it. This is just an ergonomics API so that basically just us TileDB devs can write examples more easily. I didn't originally think someone would prefer to create a String instead of just writing `&["foo", "bar"]`, but looking again I could see it.

tiledb/api/src/range.rs
davisp: Not at all. Though I tend to start implementing from the point of view of general principles. I.e.,, "I already have to write X to do Y, how hard is it to make it also do Z, Q, R". I figured, well, attributes have all the things for these checks and we have the field API so why not just `check_field_compatibility` at first. Then after I had it mostly written and was checking core for extra limitations I might be missing I realized that wasn't as general so I left a note.

tiledb/api/src/range.rs
davisp: AFAIK, its not that it'd be difficult to implement, but that it opens up a bunch of worm cans with respect to the semantics and what not.

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+
+        match cell_val_num {
+            CellValNum::Fixed(cvn) => {
+                if cvn.get() > 1 {
+                    return Err(Error::InvalidArgument(anyhow!(
+                        "Invalid cell val number: {}",
+                        cvn.get()
+                    )));
+                }
+                if datatype == Datatype::StringAscii {
```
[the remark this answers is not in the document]
davisp: Narps. We don't have a unicode dependency to handle the sort semantics.

tiledb/api/src/range.rs
davisp: Will do. I've actually already picked that approach up from reading your examples and once I spoke macro rules fluently enough it became natural to mimic.

tiledb/api/src/range.rs
```diff
@@ -0,0 +1,841 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::collection::vec;
+    use proptest::prelude::*;
+
+    // These tests are a bit awkward in that we're testing a bunch of different
+    // things in a given test. This approach was motivated by looking at
+    // llvm-cov reports and each fn_typed! generates significant overhead when
+    // compiling for coverage. Hence, one per type of Range and then just
```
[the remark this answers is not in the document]
davisp: Good idea. I might even be able to cut out some redundancy between tests with generic test functions.

tiledb/api/src/array/dimension/mod.rs
davisp: Actually, are you ok if I file a follow up story for this bit? I went to start implementing this and realized that it'll spider pretty far if I start tweaking the DimensionData class to be strongly typed. I absolutely think we should do that, but I'm thinking it s a bit much for this PR.

tiledb/api/src/array/dimension/mod.rs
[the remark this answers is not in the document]
davisp: To be clear, I mean specifically changing the DimensionData domain and extent fields from serde_json values into strongly typed Range values.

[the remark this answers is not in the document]
davisp: I replied to the wrong comment... le sigh.

tiledb/api/src/array/dimension/mod.rs
davisp: To be clear, I mean specifically changing the DimensionData domain and extent fields from serde_json values into strongly typed Range values. (Copying from my mixup. Leaving the original to avoid confusion which this probably only adding to).

rroelke: Follow-up story sounds good, it's tricky to get things right with all the corner cases. Oof.

tiledb/api/src/array/schema/mod.rs
rroelke: That smells like an Option but without the strong semantics, doesn't it? Might as well make all the `ThingData` classes have `Option<ArrayType>`.

Except not quite, because you do get the relative user-friendliness of not having to write anything down for the other fields, as you show here...

I think I like the constructor and builder the most but I like having `Dense` as the default more than I like having an `Option`, so I'll accept this change.

tiledb/api/src/range.rs
rroelke: (To be clear I do not object to the use of `Box`)

But we don't really keep pointers to the sliced thing for long, do we?  Doesn't the C API copy stuff out of whatever we gave it?

tiledb/api/src/range.rs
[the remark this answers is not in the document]
rroelke: Oh right, i18n and locales.

tiledb/api/src/range.rs
davisp: True for the most part, except for the big ones for query buffers. We've also done the thing where we re-set each of those pointers on every call to be every sure as well which is good for defense in depth.

tiledb/api/src/array/dimension/mod.rs
davisp: Filed as [sc-46692]

tiledb/api/src/array/dimension/mod.rs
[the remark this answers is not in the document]
davisp: Fixed in commit: `Fix Factor implementation for DimensionData`

tiledb/api/src/query/subarray.rs
davisp: This actually turned out more difficult than expected. It feels like it should work, but something about getting a schema from the array on a query is causing lifetime issues:

Code:
https://github.com/TileDB-Inc/tiledb-rs/commit/b697050067c4e92cc144f79dd601d41e0ef5b992

Compiler Error:
https://gist.github.com/davisp/10755108b147a004825eb76193db4a3d

I'll try fixing this later but might end up abandoning it since it might all be moot if we remove 'ctx lifetimes.

tiledb/api/examples/multi_range_subarray.rs
davisp: [Fix example description](https://github.com/TileDB-Inc/tiledb-rs/pull/72/commits/85e01c0022fb3ebf0acfaea3af0ba4c405edcc74)

tiledb/api/src/query/subarray.rs
davisp: [Check the datatype for point ranges.](https://github.com/TileDB-Inc/tiledb-rs/pull/72/commits/dbd194cf69f6a423cd84f8eb65efc77381a91eb0)

tiledb/api/src/range.rs
davisp: [Add zero copy array VarValueRange helper](https://github.com/TileDB-Inc/tiledb-rs/pull/72/commits/854614b19fa7ead0a0892304c0ffd66ef976c2f4)

tiledb/api/src/query/subarray.rs
davisp: [Add tests for Subarray::ranges()](https://github.com/TileDB-Inc/tiledb-rs/pull/72/commits/9a5186e97daee6201cb3a4b64b0bb66f4c115c35)

tiledb/api/src/query/subarray.rs
davisp: [I get by with a little help from my friends.](https://github.com/TileDB-Inc/tiledb-rs/pull/72/commits/f4cbf19767080134d533506d18b58067622c41b6)

Thanks for the assist!

tiledb/api/src/array/dimension/mod.rs
rroelke: Looks like you can delete `new_string` and just use `new_raw`?

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -206,6 +205,64 @@ impl<'ctx> Builder<'ctx> {
...
+        })?;
+        Ok(Builder {
+            dim: Dimension {
+                context,
+                raw: RawDimension::Owned(c_dimension),
+            },
+        })
+    }
+
+    // This is internal only on purpose.
```
rroelke, line 236: The comment should probably include some notes from the above discussion as to what's up

tiledb/api/src/range.rs
rroelke: How come you didn't just use `CellValNum`? That would be a lot clearer

tiledb/api/src/array/dimension/mod.rs
davisp: Fixed.

tiledb/api/src/range.rs
davisp: Totes just saw this in my email so I'm un-resolving this thread. We should probably codify a better protocol for when to resolve an open thread like this rather than relying on timeouts like I did here.

To answer the question though, I basically hadn't internalized the `CellValNum` type at the time I wrote this. I was relying on my knowledge of internals rather than preventing the edge cases. As in, I was more worried about user experience with syntax than forcing correctness by construction.

I'll go back and change this to something better.

Also, it occurs to me that `CellValNum` only has the `Fixed` and `Var` variants. I'm pretty sure it should have `Single` `Multi` and `Var` variants to correctly express the current constraints. That'll probably be a decent change.

rroelke: Reasons favoring `CellValNum::Single`:
- clarifies some pattern matching,`CellValNum::Fixed(nz) if nz.get() == 1` could be changed to `CellValNum::Single`
- construction of `CellValNum::Fixed(1)` is less work

Reasons against:
- Some code doesn't care whether `nz.get() == 1` and this code will get slightly worse
- To prevent overlap between `CellValNum::Single` and `CellValNum::Multi(1)` we would probably need to invent a new type `BoundedInt<2, u32::MAX - 1>` or something. I don't hate that idea and of course there is already a crate for it.

At some point in a branch I added `fn single() -> Self` which addresses the construction problem. So I feel sort of neutrally about this notion.

But regardless I look forward to seeing this changed to `CellValNum`

tiledb/api/src/array/dimension/mod.rs
rroelke: From slack discussion -

The current difference is that `new_string` is `pub` so we need both to export the functionality to the example.  And the linked story is the plan to straighten this out so that further than that.

tiledb/api/src/query/mod.rs
```diff
@@ -46,6 +47,20 @@ pub trait Query<'ctx> {
     fn finalize(self) -> TileDBResult<Array<'ctx>>
     where
         Self: Sized;
+
+    fn subarray(&'ctx self) -> TileDBResult<Subarray<'ctx>> {
```
rroelke, line 51: I don't love the overload of the `'ctx` lifetime name, I'd suggest `'query`

rroelke:
Good stuff, I'll approve after the last two tweaks :)

davisp:
I assume one is the 'ctx -> 'query which is fine, but what's the other tweak?

rroelke:
> I assume one is the 'ctx -> 'query which is fine, but what's the other tweak?

Using `CellValNum`

tiledb/api/src/query/mod.rs
davisp: Actually, it turns out to be not fine. I was pretty sure that `'ctx` was more correct here when I started, but who cares about naming. But now I've had rustc tell me "No" enough times that I'm pretty sure I understand specifically why (except one particular detail I need to go find an answer for).

Regardless, I believe the `'ctx` is correct here because its referring to the Schema reference in the Subarray struct which flows from the same `'ctx` lifetime that parameterizes the Query instance. What you might be thinking of is the lifetime of the Subarray struct instance itself, which is the elided lifetime of the Query instance. This can be confirmed you insert a `drop(query)` after getting the Subarray and then attempting to use the subarray.

The *only* thing I couldn't figure out is why a restricted lifetime wouldn't work. I couldn't decide if this was because the compiler tracks enough to realize that the lifetime parameter for the Query must match the lifetime parameter of the Schema (regardless of what they're called) or if there's a weird issue with restricted lifetimes on generic trait methods. But all I could get in the attempt was for the rustc to tell me that `'ctx: 'query` is shadowing an already defined lifetime `'ctx`. I'm still reading on that one.

rroelke (approved):
thanks for all the good discussions!

PR #74, opened 2024-04-25
impl Send for Array by consuming the tiledb Context

rroelke:
This test does `impl Send for Array` by having Array consume and own the `tiledb::Context`.  Array was not previously `Send` because the reference member `&'ctx Context` is only `Send` if `Context` is `Sync`.

This implementation is likely the quickest and dirtiest way to proceed but produces some usability warts which are very visible in the examples, so I think it's worth spending a little time exploring other possible ways to address the need for `impl Send for Array`.


rroelke:
Closing this because I don't like it.  If we generically need `Send` then there should be better ways to do it.

PR #75, opened 2024-04-25
Make Array Send by using `Arc<Mutext<RawContext>>`

davisp:
This makes Array Send by putting the RawContext into an `Arc<Mutext<T>>`.

### DIFF
--- a/tiledb/api/src/context.rs
+++ b/tiledb/api/src/context.rs
@@ -1,3 +1,5 @@
+use std::sync::{Arc, Mutex};
+
 use crate::config::{Config, RawConfig};
 use crate::error::{Error, RawError};
 use crate::filesystem::Filesystem;
@@ -12,6 +14,8 @@ pub enum ObjectType {
 pub(crate) struct RawContext {
     raw: *mut ffi::tiledb_ctx_t,
 }
+
+unsafe impl Send for RawContext {}
 
 impl Drop for RawContext {
     fn drop(&mut self) {
@@ -50,7 +54,7 @@ where
 }
 
 pub struct Context {
-    raw: RawContext,
+    raw: Arc<Mutex<RawContext>>,
 }
 
 impl Context {
@@ -64,7 +68,7 @@ impl Context {
         let res = unsafe { ffi::tiledb_ctx_alloc(cfg.capi(), &mut c_ctx) };
         if res == ffi::TILEDB_OK {
             Ok(Context {
-                raw: RawContext { raw: c_ctx },
+                raw: Arc::new(Mutex::new(RawContext { raw: c_ctx })),
             })
         } else {
             Err(Error::LibTileDB(String::from("Could not create context")))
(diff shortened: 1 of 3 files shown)


tiledb/api/src/context.rs
```diff
@@ -68,95 +68,110 @@ impl Context {
...
-            )
+    pub fn capi_call<Callable>(&self, action: Callable) -> TileDBResult<()>
+    where
+        Callable: FnOnce(*mut ffi::tiledb_ctx_t) -> i32,
+    {
+        // Scope the lock so we can get the error if one happens.
+        let res = {
+            // Docs say that the error result from a poisoned lock are usually
+            // just propagated to all threads by just calling unwrap on locks.
+            let raw_ctx = self.raw.lock().unwrap();
```
rroelke, line 86: Per my above comment I think this should be reverted, but...

It is also done incorrectly as the lock must be held before `action` begins and released after `expect_last_error` ends.

tiledb/api/src/context.rs
```diff
@@ -68,95 +68,110 @@ impl Context {
...
+            ffi::tiledb_ctx_get_config(ctx, &mut c_cfg)
+        })?;
+
+        Ok(Config {
+            raw: RawConfig::Owned(c_cfg),
+        })
     }
 
-    pub fn get_last_error(&self) -> Option<Error> {
+    // This is a bit awkard. We *expect* this to return an Err variant of the
```
rroelke, line 123: Are your concerns allayed if these are changed to `pub(crate)`?

All the API calls should check these, there should be no reason for a user to have to do so.

tiledb/api/src/context.rs
```diff
@@ -68,95 +68,110 @@ impl Context {
...
-            false
-        }
+        self.capi_call(|ctx| unsafe {
+            ffi::tiledb_ctx_is_supported_fs(ctx, fs as u32, &mut supported)
+        })?;
+
+        Ok(supported == 1)
     }
 
     pub fn set_tag(&self, key: &str, val: &str) -> TileDBResult<()> {
```
rroelke, line 161: Based on the discussion in `#core` the other day, this should be required as part of the context construction? Should we file a story for that?

tiledb/api/src/context.rs
davisp: Ohhhhh, good catch!

tiledb/api/src/context.rs
davisp: I was mostly just not liking my weird use of TileDBResult here. It was just "Wait, should an error be returned as an `Ok` or an `Err`?" sort of weirdness. As you pointed out we'd want to make these called while the lock is held if we go that direction.

tiledb/api/src/context.rs
davisp: Done: https://app.shortcut.com/tiledb-inc/story/46092

tiledb/api/src/context.rs
```diff
@@ -37,27 +32,32 @@ pub trait ContextBound<'ctx> {
...
+    where
+        Callable: FnOnce(*mut ffi::tiledb_ctx_t) -> i32,
+    {
+        self.context().capi_call(action)
     }
 }
 
 pub struct Context {
-    raw: RawContext,
+    raw: Arc<Mutex<RawContext>>,
```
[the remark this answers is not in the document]
davisp: capi_call is now its own PR: #76

tiledb/api/src/context.rs
davisp: Fixed.

tiledb/api/src/context.rs
davisp: I did this differently and just swallowed the error like we were before so its same as it used to be.

davisp:
Closing this as we'll not likely need it.

PR #76, opened 2024-04-25
Replace capi_return with capi_call

davisp:
This removes all access to the underlying raw context pointer except through the `Context::capi_call` method. This method accepts a single argument that is a `FnOnce` callable. This callable must take a single argument which is a `*mut ffi::tiledb_ctx_t` that can be used to pass to all of the FFI calls that require it. The return value should be the i32 result of invoking one of the tiledb FFI functions.

A simple example looks like such:

```rust
pub fn version(&self) -> TileDBResult<i64> {
    let mut c_version: std::os::raw::c_int = out_ptr!();
    self.capi_call(|ctx| unsafe {
        ffi::tiledb_array_schema_get_allows_dups(
            ctx, c_schema, &mut c_version
        )
    })?;

    Ok(c_version as i64)
}
```

### DIFF
--- a/tiledb/api/src/array/attribute/arrow.rs
+++ b/tiledb/api/src/array/attribute/arrow.rs
@@ -89,7 +89,7 @@ pub fn arrow_field(
                     )
                 })?;
         Ok(Some(
-            arrow_schema::Field::new(name, arrow_dt, attr.is_nullable())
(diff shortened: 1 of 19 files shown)


davisp:
Leaving as a draft until after I update this after groups are merged.

rroelke (approved):

rroelke (approved):

PR #77, opened 2024-04-26
Implement conversion from RawReadOutput to Arrow array

rroelke:
Proptest is used for testing.

RawReadOutput could be either a primitive array or a generic list array.  Arrow `RecordBatch` is just looking for `Arc<dyn Array>`, so we use that as the arrow array output type.

### DIFF
--- a/tiledb/api/src/query/read/callback.rs
+++ b/tiledb/api/src/query/read/callback.rs
@@ -442,11 +442,15 @@ macro_rules! query_read_callback {
 
                         let [< l_ $U:snake >] = self.[< arg_ $U:snake >].location.borrow();
                         let [< input_ $U:snake >] = [< l_ $U:snake >].as_shared();
+                            /*
+                             * TODO: if it is the final result, enable moving out of
+                             * this so that we can avoid a copy
+                             */
 
                         let [< arg_ $U:snake >] = RawReadOutput {
                             nvalues: [< nvalues_ $U:snake >],
                             nbytes: [< nbytes_ $U:snake >],
-                            input: &[< input_ $U:snake >]
+                            input: [< input_ $U:snake >]
                         };
                     )+
                 }
@@ -766,7 +770,7 @@ mod tests {
             let arg = RawReadOutput {
                 nvalues: ncells,
                 nbytes: ncells * std::mem::size_of::<u64>(),
-                input: &input_data,
+                input: input_data,
             };
 
             let prev_len = unitdst.len();
@@ -891,7 +895,7 @@ mod tests {
             let arg = RawReadOutput {
                 nvalues,
                 nbytes,
-                input: &input,
+                input,
             };
             stringdst
                 .intermediate_result(arg)
--- a/tiledb/api/src/query/read/callback.rs
+++ b/tiledb/api/src/query/read/callback.rs
@@ -60,12 +60,12 @@ pub trait ReadCallbackVarArg: Sized {
 
     fn intermediate_result(
         &mut self,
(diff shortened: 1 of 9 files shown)


tiledb/api/src/query/read/callback.rs
```diff
@@ -442,11 +442,15 @@ macro_rules! query_read_callback {
...
+                            /*
+                             * TODO: if it is the final result, enable moving out of
+                             * this so that we can avoid a copy
+                             */
 
                         let [< arg_ $U:snake >] = RawReadOutput {
                             nvalues: [< nvalues_ $U:snake >],
                             nbytes: [< nbytes_ $U:snake >],
-                            input: &[< input_ $U:snake >]
+                            input: [< input_ $U:snake >]
```
davisp, line 453: Is it just me, or is paste just a fancy `##` symbol concatenation thing?

tiledb/api/src/query/read/output/arrow.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        let mut v_offsets = if let Some(offsets) = offsets {
+            match offsets {
+                Buffer::Empty => vec![],
+                Buffer::Owned(offsets) => {
+                    let mut offsets = offsets.into_vec();
+                    offsets.truncate(value.nvalues);
+                    offsets
+                }
+                Buffer::Borrowed(offsets) => offsets[0..value.nvalues].to_vec(),
```
davisp, line 91: I've got no idea how often this is called, but this strikes me as wrong by design. There shouldn't be allocations involved here. These can and will end up being gigabyte sized allocations when users eventually start banging on this. I don't think we need to change it now, but long term we'll definitely need to revisit this.

tiledb/api/src/query/read/output/arrow.rs
```diff
@@ -0,0 +1,340 @@
...
+            let gl = arrow
+                .as_any()
+                .downcast_ref::<GenericListArray<i64>>()
+                .unwrap();
+
+            let arrow_offsets = gl.offsets();
+            assert!(arrow_offsets.len() <= offsets.len() + 1);
+            assert_eq!(arrow_offsets.len(), rr.nvalues + 1);
+
+            /* arrow offsets are value, tiledb offsets are bytes */
```
davisp, line 208: Ooof. I can't wait to deal with all the bugs this difference is gonna cause.

davisp (approved):
+1. I agree that lifetime issue needs fixing but I don't think it has to happen as part of this PR.

tiledb/api/src/query/read/callback.rs
rroelke: Pretty much yep. Except normal rust macros don't let you do that.

tiledb/api/src/query/read/output/arrow.rs
rroelke: The good news is that Isaiah said we can configure TileDB to produce offsets which are records, not bytes. Might not even be a bad plan to do that as the default.

tiledb/api/src/query/read/output/arrow.rs
rroelke: If the user owns their query buffer, then this is unavoidable (without being unsafe).  Arrow either takes ownership of the buffer (as we see just above this), or copies.

If the user moves their buffer into the query, which we don't have an option for yet, or if they let tiledb manage the buffer, then we could end up with `Buffer::Owned` and avoid the copy.  I don't think that's actually implemented yet, but it is at least a possibility.

PR #78, opened 2024-04-29
Fix nightly clippy issues

rroelke:
This does not address the inevitable `proptest-derive` issue that I see when running clippy with nightly, but for some reason that did not appear in the CI output.


rroelke:
Manually triggered nightly passed:
https://github.com/TileDB-Inc/tiledb-rs/actions/runs/8877805921

PR #79, opened 2024-05-01
Fixed cell val num should be available in RawReadOutput, make it so with `CellStructure` enum

rroelke:
Currently the argument to the query callback adapter does not have the cell val num anywhere within. If the cell val num is Var, then there are offsets available; but if the cell val num is a constant other than one, then you are stuck if you need it.

We need the cell val num for our implementation of impl From<RawReadOutput> for Arc<dyn ArrowArray> so that we can properly create arrow FixedSizeList fields.

We also would benefit from the cell val num being available to make another iterator over a RawReadOutput results which yields the fixed-size records.

This pull request satisfies the above by replacing `cell_offsets: Option<Bufffer<'data, u64>>` in `QueryBuffers` and its relatives with `cell_structure: CellStructure<'data>`.  `CellStructure` is very similar to `CellValNum`, except it carries the offset data along with it when the cell val num is `Var`.

I chose this implementation rather than adding a `cell_val_num` field to affected places so that the cell structure would not be essentially duplicated by the hypothetical field and the offsets buffer. This representation prevents an invalid pairing of a `Some` offsets bufffer with a non-`Var` cell val num.


davisp (approved):
+1. Only thing that tripped me up was calling it `cell_structure` but I honestly can't think of anything better.

PR #81, opened 2024-05-02
Configure queries to return arrow-like results, i.e. "extra" offset and element offsets; remove RawReadOutput::nbytes

rroelke:
Our primary use case requires us to manifest arrow record batches.  Core has an option to produce arrow-like offsets, so let's use those options instead of re-rolling them ourselves in the Rust layer.

But even on top of that, it makes a lot of sense with the current API design to use element-based offsets.  The buffers we attach to the query API express themselves in terms of a `Unit` type which is some primitive.  By using the element offsets we save the effort from doing arithmetic from the returned query sizes.

The `extra_offset` configuration also enables us to eliminate the `RawReadOutput::nbytes` field, which is now redundant with the last offset.  This change makes the `RawReadOutput` easier to use.

### DIFF
--- a/tiledb/api/src/query/read/output/mod.rs
+++ b/tiledb/api/src/query/read/output/mod.rs
@@ -22,7 +22,6 @@ pub mod strategy;
 
 pub struct RawReadOutput<'data, C> {
     pub ncells: usize,
-    pub nbytes: usize,
     pub input: QueryBuffers<'data, C>,
 }
 
@@ -30,8 +29,19 @@ impl<C> RawReadOutput<'_, C> {
     pub fn nvalues(&self) -> usize {
         match self.input.cell_structure {
             CellStructure::Fixed(nz) => self.ncells * nz.get() as usize,
-            CellStructure::Var(_) => self.nbytes / std::mem::size_of::<C>(),
+            CellStructure::Var(ref offsets) => {
+                if offsets.is_empty() {
+                    0
+                } else {
+                    assert!(self.ncells < offsets.len());
+                    offsets.as_ref()[self.ncells] as usize
+                }
+            }
         }
+    }
+
+    pub fn nbytes(&self) -> usize {
+        self.nvalues() * std::mem::size_of::<C>()
     }
 
     fn to_json(&self) -> serde_json::value::Value
@@ -57,7 +67,7 @@ impl<C> RawReadOutput<'_, C> {
 
         json!({
             "ncells": self.ncells,
-            "nbytes": self.nbytes,
+            "nbytes": self.nbytes(),
             "data": {
                 "capacity": self.input.data.len(),
                 "defined": self.nvalues(),
@@ -81,7 +91,6 @@ where
 pub struct TypedRawReadOutput<'data> {
     pub datatype: Datatype,
     pub ncells: usize,
-    pub nbytes: usize,
     pub buffers: TypedQueryBuffers<'data>,
 }
 
@@ -93,7 +102,6 @@ impl<'data> TypedRawReadOutput<'data> {
         TypedRawReadOutput {
             datatype,
             ncells: rr.ncells,
-            nbytes: rr.nbytes,
             buffers: rr.input.into(),
         }
     }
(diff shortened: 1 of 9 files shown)


rroelke:
**Note** this pull request cannot be merged, all the tests pass but the examples do not yet pass!

I am heading out for the day and am creating this now in case @davisp you had a chunk of time to review it today. Hopefully it will not drift much as I debug the examples.

tiledb/api/src/query/read/output/strategy.rs
```diff
@@ -193,32 +193,53 @@ where
...
+                         * What does "empty" look like for arrow-shaped offsets?
+                         * Is it an empty offset buffer?
+                         * Or is it a single offset `[0]`?
+                         * The arrow `OffsetBuffer` doc suggests the latter, but is the tiledb core
+                         * implementation compliant with that?
+                         * A glance at the source code suggests it is not, i.e. if the number
+                         * of records is zero then the offsets will be empty.
+                         * So we must be able to generate empty offsets here too, and
+                         * upstream code must be ready.
+                         */
```
davisp, line 208: The config name of "extra_offset" suggests to me that the offsets for var data should be the moral equivalent of `&[0u64]` and if its not that, then we should probably file a bug against core if that's what Arrow expects.

davisp (approved):
+1

My only major thought is that variable name really makes me thing we should go with the `num_$thing` pattern instead.

tiledb/api/src/query/read/output/mod.rs
```diff
@@ -697,14 +701,10 @@ impl<'data, C> Iterator for VarDataIterator<'data, C> {
         let s = self.offset_cursor;
         self.offset_cursor += 1;
 
-        if s + 1 < self.ncells {
+        if s < self.ncells {
             let start = offset_buffer[s] as usize;
             let slen = offset_buffer[s + 1] as usize - start;
             Some(&data_buffer[start..start + slen])
```
davisp, line 707: I started poking at the examples to see if I couldn't debug the issue and stumbled over this bit. Unless my math ain't mathing right, this should just be:

```rust
let start = offsets_buffers[s] as usize;
let end = offsets_buffers[s + 1] as usize;
Some(&data_buffer[start..end]);
```

davisp:
I started debugging the `query_condition_dense` and got as far as finding where the logic is broken:

https://github.com/TileDB-Inc/tiledb-rs/blob/e7cc656bc47b8b0e1215e00b8237d4bad2d09e1c/tiledb/api/src/query/read/raw.rs#L263

The offsets size there is in bytes which is 88 which is correct. There's 10 elements in the array, with the extra, that's 11 u64 values so 88 bytes. So returning that we have 87 cells is not correct there. Reading the code its not entirely clear to me what the correct way is to fix that though. So I'll let you have at it in the morning.

tiledb/api/src/query/read/output/strategy.rs
rroelke: I definitely agree, it does need some verification but I'll make sure to follow up on that.

tiledb/api/src/query/read/output/mod.rs
rroelke: Probably the same to the compiler but that does look simpler to the human. I'll fix it up

rroelke:
> I started debugging the `query_condition_dense` and got as far as finding where the logic is broken:
> 
> https://github.com/TileDB-Inc/tiledb-rs/blob/e7cc656bc47b8b0e1215e00b8237d4bad2d09e1c/tiledb/api/src/query/read/raw.rs#L263
> 
> The offsets size there is in bytes which is 88 which is correct. There's 10 elements in the array, with the extra, that's 11 u64 values so 88 bytes. So returning that we have 87 cells is not correct there. Reading the code its not entirely clear to me what the correct way is to fix that though. So I'll let you have at it in the morning.

THanks for taking a stab at it.  Turns out it is pretty innocuous, just that the *sizes* returned by the read are still in bytes. We didn't configure those! And I'm not sure whether we can. Just needs some tweaks to `last_read_ncells` to fix.

PR #82, opened 2024-05-03
[sc-44537] Unified way of attaching user data to enumeration, dimension, query buffers, etc

rroelke:
This branch replaces our existing ad-hoc generic data type traits with two new traits, `PhysicalType` and `LogicalType`.  The former is meant to be a catch-all trait which encapsulates all the operations we might want to do with the stuff exchanged in buffers between the Rust and C APIs (e.g. query data, etc).  The latter is not as clear yet how it will be used but I expect it to be helpful for the work with Arrow.

This removes several of the previous ad hoc data type traits, hence my choice to use this to resolve the linked story.

Story details: https://app.shortcut.com/tiledb-inc/story/44537


tiledb/api/src/datatype/logical.rs
```diff
@@ -0,0 +1,143 @@
...
+pub trait LogicalType: crate::private::Sealed {
+    const DATA_TYPE: Datatype;
+
+    type PhysicalType: PhysicalType;
+}
+
+pub struct UInt8Type {}
+
+impl LogicalType for UInt8Type {
+    const DATA_TYPE: Datatype = <u8 as PhysicalType>::DATA_TYPE;
```
davisp, line 16: Not necessarily asking for a change here, but I will say this pattern took me more than 30s to figure out. On the one hand, it seems like just writing out `DataType::UInt8` would be clearer, though on the other hand, there's a vague "But then it's possible to get things mismatched". Personally, I'd lean towards "this is write once code that'll ~never change, only possibly be replaced wholesale" so I'd go with the `Datatype::$variant` to make things clearer.

Granted, that's only a very loosely held slight preference so feel free to ignore this.

tiledb/api/src/datatype/logical.rs
```diff
@@ -0,0 +1,143 @@
+use crate::datatype::physical::PhysicalType;
+use crate::datatype::Datatype;
+use crate::private::sealed;
+
+/// Trait which provides statically-typed attributes for a TileDB `Datatype`
+/// for use with generics.
+pub trait LogicalType: crate::private::Sealed {
```
davisp, line 7: ~~How is this possible? That blog post I linked in a different comment uses this specifically as a "you can't do this" so I'm surprised it compiles. I'm assuming that I'm missing a privacy specifier else where, but if so then we might want to tag this with a comment about the how part.~~

Derps again. Its using the first pattern after the bad pattern where the Sealed trait is public inside a private module. Which... huh. Perhaps that's a thing? Actually, does that mean our use of `pub mod` everywhere is probably wrong? Something like, everything module should be private and then we go through and re-export public symbols with clear intention? I did mention the other day we should clean up some of our exports and that feels like the sort of approach that'd make things clear. Obviously that's not part of this story. Just something to think about.

tiledb/api/src/datatype/logical.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+pub struct StringUcs4Type {}
+
+impl LogicalType for StringUcs4Type {
+    const DATA_TYPE: Datatype = Datatype::StringUcs4;
+    type PhysicalType = u32;
+}
+
+sealed!(UInt8Type, UInt16Type, UInt32Type, UInt64Type);
```
davisp, line 133: ~~If `LogicalType` is already constrained by `crate::private::Sealed` in the trait definition, why does it need repeating here?~~

Derps. These are implementing the Sealed trait for the type so that they meet the conditions of the constraint for `LogicalType`.

davisp (approved):
+1. I also went and looked and noticed that [ArrowNativeType is also a sealed](https://docs.rs/datafusion/latest/datafusion/common/arrow/datatypes/trait.ArrowNativeType.html#sealed) trait which is more than enough convincing that this is a solid approach. I'd just ask to add a bit extra context on the sealed trait to save future folks a bit of head scratching on the intent there.

tiledb/api/src/lib.rs
```diff
@@ -79,3 +78,19 @@ pub trait Factory<'ctx> {
 
     fn create(&self, context: &'ctx context::Context) -> Result<Self::Item>;
 }
+
+mod private {
+    // the "sealed trait" pattern prevents downstream crates from implementing a trait
+    // if you make `Sealed` a super-trait. This is important for something like `PhysicalType`.
+    pub trait Sealed {}
```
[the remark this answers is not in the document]
rroelke: Yeah that's totally fair, I'll clarify the motivation.  Yep definitely want to avoid surprises if someone makes a `u256` or something and is like "tiledb should be able to use thhat!"

tiledb/api/src/datatype/logical.rs
rroelke: I was kind of unsure if the `PhysicalType` trait should have a Datatype associated constant at all, so maybe I'll remove that and follow along with this.

davisp: Fair point. I could maybe see there being a "default logical type", though given my assumption that most uses of our API bindings will be "programmatic" instead of user written so the ergonomics probably won't matter there.

tiledb/api/src/lib.rs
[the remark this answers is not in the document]
davisp: Excellent. Thanks!

tiledb/api/src/datatype/mod.rs
```diff
@@ -517,211 +523,184 @@ impl TryFrom<ffi::tiledb_datatype_t> for Datatype {
...
-            Datatype::GeometryWkt => unimplemented!(),
-        }
-    }};
-
-    ($func:ident, $datatype:expr$(, $arg:expr)* => $then:expr) => {{
         type Datatype = $crate::Datatype;
         match $datatype {
             Datatype::Int8 => {
-                let $func = $func::<i8>($($arg,)*);
+                type $typename = $crate::datatype::Int8Type;
```
davisp, line 529: This all looks great, though I would have a slight preference for using fully qualified names instead of the re-exports. Feel free to not bother here though. It'll be a straightforward mechanical change if/when we get around to clarifying our public symbols and so on.

rroelke: You're definitely right about this, updated.

PR #83, opened 2024-05-04
Remove the 'ctx lifetime

davisp:
In hindsight, this is a shining example of where we should use an
Rc<RefCell<T>> as it exactly matches the semantics of what we want for
Context's. There's a decent bit of churn here, but its all mechanical
and basically just boils down to removing every occurrence of 'ctx in
the code base.


tiledb/api/src/context.rs
```diff
@@ -46,8 +48,21 @@ impl Drop for RawContext {
     }
 }
 
-pub trait ContextBound<'ctx> {
-    fn context(&self) -> &'ctx Context;
+// I tried using this to mark all the types that just return a self.context
+// vs the types that return self.member.context(). It all compiled until I
+// tried implementing CApiInterface trait for it which caused a trait
+// implementation conflict I couldn't figure out. I'm leaving this and all
+// the marker traits commented out just so we have something to search for
```
rroelke, line 55: To be honest I'm kind of surprised that this made it as far as it did.
I'm sure the compiler would eventually have complained that `Self` doesn't necessarily have a `context` field.  It's kind of a form of specialization.

That is one downside with the compilation errors coming in stages, you can work really hard clearing out the errors from one stage only to have the next come around and give you another 100 errors

tiledb/api/src/context.rs
```diff
@@ -72,8 +87,9 @@ where
     }
 }
 
+#[derive(Clone)]
 pub struct Context {
-    raw: RawContext,
+    raw: Rc<RefCell<RawContext>>,
```
rroelke, line 92: Why do you need the RefCell?

rroelke:
I haven't really invested the time yet into determining if this is the *right* thing to do, but I do think it's better than before because it will help to iterate slightly faster, and doesn't introduce any new problems.

I'm not sure it was necessary to remove the proc macros - they would have become a lot simpler without the need to deal with the lifetimes - but they were only ever a way to theoretically iterate faster as the query API was evolving, and that isn't as important now.

Not approving yet - see other comments.

tiledb/api/src/context.rs
davisp: Oooh, I need to delete that. The entire point for going this direction in the first place was to guarantee that we returned called clone() when getting down to a `ContextBound` implementation that has an actual `Context` member. But given that the trait now returns `Context` instead of `&Context` it's a compiler error to not return `self.context.clone()` which is all I wanted.

tiledb/api/src/context.rs
davisp: I probably don't in hindsight. I just did `Rc<RefCell<T>>` because that's basically used in all the places. I'll try and remove it and see if we can get away without it.

rroelke: I'm sure you can.  A Rust-native thing would need it for interior mutability, but we get that via the `ffi`

tiledb/api/src/context.rs
rroelke: > to not return self.context.clone() which is all I wanted

I think the recommendation is to qualify it a la `Rc::clone(&self.context)` which is functionally not different but tells the reader that it's not an expensive clone.  I don't remember where I read that.

davisp:
> I'm not sure it was necessary to remove the proc macros

See the commit message here: https://github.com/TileDB-Inc/tiledb-rs/pull/83/commits/0a3bcd197fa0d2bde904210d3fbe1361a15ded92

Bottom line is I had to remove them to be able to understand what they were doing. After removing them I then realized that it resulted in a net removal of 465 lines so I didn't feel like adding code back that would just increase the line count. The ContextBound macro was the worst of the two by a lot. The implementation and tests were around 500 LoC which means we'd have to have 167 definitions of `impl ContextBound` to break even there. And those definitions are trivial to read and write so it just really doesn't make a whole lot of sense to keep them around. The Query trait on the other hand was mind bending. It wasn't until the very last change I made to remove them that it really clicked what was going on.

I don't really buy the premise that these would help iterate faster. Certainly not the ContextBound macro given its trivial. The Query example is the opposite in that it was hiding some complex trait implementations which is almost worse. I'm really starting to lean towards only allowing proc macros in PRs that are specific to adding the proc macro since that would show both the complexity of the implementation and the benefit we're getting from that complexity when we see how it affects already implemented code.

It just seems to me that if we're reaching for proc macros to make something easier on the first implementation, its more of a red flag that the approach to the implementation should be rethought.

tiledb/api/src/context.rs
davisp: Huh. Apparently it comes from the Rust book and there's a (disabled due to false positive) lint for this approach. Good enough for me to change to that style.

tiledb/api/src/context.rs
davisp: Oh, I'm fairly certain of it as well. Just not 100% certain that there's nothing subtle relying on a trait for RefCell that's not apparent. Easiest way to find out is to just give it a whirl which should be something like two or three lines of code.

rroelke (approved):

PR #85, opened 2024-05-05
Fix minor bugs

davisp:
I spent a day today taking another stab at learning the proptest API. So far my FilterListData Strategy is generating about 100k/sec valid cases. During that work I found a couple minor bugs in our Rust logic that contradicted core.

The first bug is a simple typo fixed by doing a `s/Boolean/Blob/`.

The second one is actually interesting. [The C++ code looks like this][1]:

```c++
bool CompressionFilter::accepts_input_datatype(Datatype input_type)  const {
    auto this_filter_type = compressor_to_filter(compressor_);
    if (this_filter_type == FilterType::FILTER_DOUBLE_DELTA ||
        this_filter_type == FilterType::FILTER_DELTA) {
      // Delta filters do not accept floating point types.
      if (datatype_is_real(
        reinterpret_datatype_ != Datatype::ANY ? reinterpret_datatype_ :
                                                 input_type)) {
        return false;
    }
  }

  return true;
}
```

The previous Rust translation of that logic [looked like this][2]:

```rust
FilterData::Compression(CompressionData { kind, .. }) => match kind
{
    CompressionType::Delta {
        reinterpret_datatype,
    }
    | CompressionType::DoubleDelta {
        reinterpret_datatype,
    } => {
        // these filters do not accept floating point
        let check_type =
            if let Some(Datatype::Any) = reinterpret_datatype {
                *input
            } else if let Some(reinterpret_datatype) =
                reinterpret_datatype
            {
                reinterpret_datatype
            } else {
                return None;
            };
        if check_type.is_real_type() {
            None
        } else {
            Some(check_type)
        }
    }
    _ => Some(*input),
},
```

And the corrected translation looks like this:

```rust
FilterData::Compression(CompressionData { kind, .. }) => match kind
{
    CompressionType::Delta {
        reinterpret_datatype,
    }
    | CompressionType::DoubleDelta {
        reinterpret_datatype,
    } => reinterpret_datatype.map_or(Some(*input), |dtype| {
        if !dtype.is_real_type() {
            Some(dtype)
        } else {
            None
        }
    }),
    _ => Some(*input),
},
```

That's fairly subtle.

The reason the first Rust version is wrong is because it misses the one of four cases where reinterpret_datatype is None and the input type is valid for the filter which doesn't happen very often given the specificity of the input datatype.

I love proptets.

[1]: https://github.com/TileDB-Inc/TileDB/blob/d00866b64eab47929170fe61f3498319d7b9711d/tiledb/sm/filter/compression_filter.cc#L98-L111
[2]: https://github.com/TileDB-Inc/tiledb-rs/blob/94cdfa59064548e3e43bc77265c5022324ae73ad/tiledb/api/src/filter/mod.rs#L230-L248

### DIFF
--- a/tiledb/api/src/datatype/mod.rs
+++ b/tiledb/api/src/datatype/mod.rs
@@ -366,7 +366,7 @@ impl Datatype {
     pub fn is_byte_type(&self) -> bool {
         matches!(
             *self,
-            Datatype::Boolean | Datatype::GeometryWkb | Datatype::GeometryWkt
+            Datatype::Blob | Datatype::GeometryWkb | Datatype::GeometryWkt
         )
     }
 
--- a/tiledb/api/src/filter/mod.rs
+++ b/tiledb/api/src/filter/mod.rs
@@ -223,24 +223,13 @@ impl FilterData {
                 }
                 | CompressionType::DoubleDelta {
                     reinterpret_datatype,
-                } => {
-                    // these filters do not accept floating point
-                    let check_type =
-                        if let Some(Datatype::Any) = reinterpret_datatype {
-                            *input
-                        } else if let Some(reinterpret_datatype) =
-                            reinterpret_datatype
-                        {
-                            reinterpret_datatype
-                        } else {
-                            return None;
-                        };
-                    if check_type.is_real_type() {
-                        None
+                } => reinterpret_datatype.map_or(Some(*input), |dtype| {
+                    if !dtype.is_real_type() {
+                        Some(dtype)
                     } else {
-                        Some(check_type)
+                        None
                     }
-                }
+                }),
                 _ => Some(*input),
             },
             FilterData::ScaleFloat { byte_width, .. } => {
--- a/tiledb/api/src/filter/mod.rs
+++ b/tiledb/api/src/filter/mod.rs
@@ -223,13 +223,40 @@ impl FilterData {
                 }
                 | CompressionType::DoubleDelta {
                     reinterpret_datatype,
-                } => reinterpret_datatype.map_or(Some(*input), |dtype| {
-                    if !dtype.is_real_type() {
-                        Some(dtype)
-                    } else {
-                        None
+                } => {
+                    // Delta and Double Delta filters don't take float types
+                    // as input. However, it also has the ability to reinterpret
+                    // any datatype so we have to handle that as well.
+
+                    // If reinterpret_datatype is None, then we just check
+                    // the input datatype.
+                    if reinterpret_datatype.is_none() && !input.is_real_type() {
+                        return Some(*input);
                     }
-                }),
+
+                    let reinterpret_datatype = reinterpret_datatype.unwrap();
+
+                    // Reinterpret datatype is set. So now we have to match the
+                    // logic in core which is that if its set to Datatype::Any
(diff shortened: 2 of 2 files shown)


tiledb/api/src/datatype/mod.rs
```diff
@@ -366,7 +366,7 @@ impl Datatype {
     pub fn is_byte_type(&self) -> bool {
         matches!(
             *self,
-            Datatype::Boolean | Datatype::GeometryWkb | Datatype::GeometryWkt
+            Datatype::Blob | Datatype::GeometryWkb | Datatype::GeometryWkt
```
rroelke, line 369: Looks like this changed in core at
```
commit 202930837f8b1ff57fcee954ca0f08009a30858c
Author: Beka Davis <31743465+bekadavis9@users.noreply.github.com>
Date:   Mon Mar 25 15:54:30 2024 -0400
```

So at the time it was written, it may well have been correct!

That's the trouble with having to replicate some of this logic from core in this layer, I suppose.

tiledb/api/src/filter/mod.rs
```diff
@@ -223,24 +223,13 @@ impl FilterData {
...
-                        {
-                            reinterpret_datatype
-                        } else {
-                            return None;
-                        };
-                    if check_type.is_real_type() {
-                        None
+                } => reinterpret_datatype.map_or(Some(*input), |dtype| {
+                    if !dtype.is_real_type() {
+                        Some(dtype)
```
rroelke, line 228: This looks to me like a regression if `dtype == Datatype::Any`, shouldn't there still be a case for that?

Or, as a philosophical matter, should we remove `Datatype::Any` from the Rust layer and change the places where it is allowed to `Option<Datatype>`?  Or should we draw inspiration from `NonZero<T>` and make a `DatatypeNotAny`?

But none of these are really right since `Any` is an actual type which requires `CellVarNum::Var`.

tiledb/api/src/datatype/mod.rs
davisp: Well it was never correct, but we missed finding a bug in core by ~12 days.

tiledb/api/src/filter/mod.rs
davisp: Well crap. I apparently managed t delete all the proptests code I wrote this weekend. Not a big deal as it was mostly just a learning exercise, though I did have a quite nice filter list generation property. I'll work later on getting that rewritten to show off.

I'll have to double check whether or not passing Any is accepted as a reinterpret_datatype or if it gets tossed in an exception. It could go both ways at multiple layers. If it does then yeah, this code is rejecting a different corner of the truth table instead.

davisp: Proptests stuff rewritten and on a branch (so I don't do another dumb and accidentally delete it). You definitely caught an edge case there. If a user sets the datatype to Any we have to handle it because TileDB accepts that.

PR is updated with the fix.

tiledb/api/src/datatype/mod.rs
rroelke: Does "correct" mean "does the right thing in principle" or "does what core does"?
:)

rroelke (approved):

PR #86, opened 2024-05-06
`typed_enum!` macro implements several step of our `TypedThing` pattern

rroelke:
We have `metadata::Value`, `TypedQueryBuffers`, and something with ranges... and maybe more?

This pull request brings that pattern into a common core of a new macro `typed_enum!` which expands into an `enum` definition with one variant per logical type, and also into a new macro definition which matches on the enum and then brings the logical type into context for an expression block (like `fn_typed`).

The choice of using logical type rather than physical type for variation was motivated by what I'm doing with arrow.

### DIFF
--- a/tiledb/api/src/datatype/mod.rs
+++ b/tiledb/api/src/datatype/mod.rs
@@ -711,6 +711,255 @@ macro_rules! fn_typed {
     }};
 }
 
+#[macro_export]
+macro_rules! typed_enum {
+    ($src:ident < $range:ident > => $dst:ident) => {
+        typed_enum!($src <() enum $range> => $dst);
+    };
+    ($src:ident < $g1:tt, $range:ident > => $dst:ident) => {
+        typed_enum!($src <($g1) enum $range> => $dst);
+    };
+    ($src:ident < $g1:tt, $g2:tt, $range:ident > => $dst:ident) => {
+        typed_enum!($src <($g1, $g2) enum $range> => $dst);
+    };
+    ($src:ident < $g1:tt, $g2:tt, $g3:tt, $range:ident > => $dst:ident) => {
+        typed_enum!($src <($g1, $g2, $g3) enum $range> => $dst);
+    };
+    ($src:ident < $g1:tt, $g2:tt, $g3:tt, $g4:tt, $range:ident > => $dst:ident) => {
+        typed_enum!($src <($g1, $g2, $g3, $g4) enum $range> => $dst);
+    };
+    ($src:ident < ($($generics:tt),*) enum $range:ident > => $dst:ident) => {
+        pub enum $dst<$($generics),*> {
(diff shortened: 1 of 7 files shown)


tiledb/api/src/datatype/mod.rs
```diff
@@ -711,6 +711,256 @@ macro_rules! fn_typed {
     }};
 }
 
+#[macro_export]
+macro_rules! typed_enum {
```
rroelke, line 715: Needs comments reflecting the PR description which I will add as this PR is discussed.

davisp (changes requested):
I'm really not a fan of this approach for a bunch of vague reasons I've spent the last two hours trying to turn into more than a list of "Enh, seems not great" descriptions.

However, eventually in that list of reasons I got to "How would we document the variants and how would users see those docs?" and it looks like `cargo doc` doesn't render/show macro stuff? If you do a before/after render and look at something like the metadata variants they just disappear from the generated docs. I was expecting to just see the blank "here's the list" sort of docs with no textual description, but to see them not appear at all is a deal breaker for public API docs.

Perhaps there's a way around that, but it's definitely the first thing that needs addressing.

rroelke:
I was actually a little relieved to see you push back on this because as much it was kind of fun to do I was also pretty leery of it by the time it got here.

I have different questions about it, though. Namely... do we actually want these enums to range on the logical type? Or is the physical type good enough?  My arrow code, as you may soon see, does present a weak argument in favor, but I'm not confident we know enough yet to understand what would get *worse*.

The macro was fun to work on but I totally agree that not having a public API is a deal-breaker as the choice of implementation.

Given that, did the choice of logical type for the enum variants also bother you for reasons that were hard to articulate?  Or do you endorse that notion, just with a different implementation?

rroelke:
https://github.com/TileDB-Inc/tiledb-rs/pull/87/files#r1592636193

This comment on the new PR is why I chose to even try this

davisp:
I was against applying it everywhere, that's for sure. I'm starting to see some vague signals (things like rust-analyzer lag increasing significantly) that are leading me to think we're getting a bit out of scope for what are supposed to be some bindings to a C API. Expanding a bunch of our variants significantly doesn't seem like a step in the right direction.

That said, doing it for one specific type here or there isn't the end of the world. We do have to handle all of those variants somehow, after all.

I'm not sure I'm catching all of the context of that comment on the other PR, but I don't think having typed buffers for all logical types is the answer. If you're just trying to get the datatype in there for context in the conversion, just change your `TryFrom<TypedRawReadOutput>` to `TryFrom<(Datatype, TypedRawReadOutput)>` or something. Though personally I would have expected something shaped more like `TypedRawReadOutput::create_array(field: ArrowField) -> ArrowArray` depending various other bits around allocated buffer management in the query that I've not thought about enough to know if getting crazy and doing zero copy buffer transfers would be possible.

PR #87, opened 2024-05-07
Fix up arrow datatype compatibility and enable turning `TypedRawReadOutput` into `arrow::RecordBatch`

rroelke:
This is a bit of a doozy.  It all started with our initial "select star" success with the tables repo... that did not work for variable-length data.  And as I would discover by working on this, it probably also does not work for anything other than primitive integer/float data.

This pull request should close that gap by providing a method for constructing a `RecordBatch` from `TypedRawReadOutput`, which is basically what the tables repo wants to do.  This was a challenge due to some data correctness issues which have been addressed in previous PRs, but also because the `RecordBatch` does a strong schema check of the logical data types when you try to construct it for a given schema.

As a result, a lot of this PR, maybe even most of it, concerns itself with strengthening our arrow datatype layer.  We fill in many more cases, and distinguish between "exact" and "inexact" conversions, which I believe will be important in the future so we could, for example, inspect tiledb geometry data from Tables by matching it to an inexact `LargeBinaryArray`.

The other half of the PR adds more capabilities to materializing arrow arrays.  We add some wrappers for `QueryBuffers` to make statically-known assertions about the contents, so that we can materialize a strongly-typed arrow array instead of a `Box<dyn ArrowArray>` trait object.  Using those primitives, we can `impl From<TypedRawReadOutput> for Arc<dyn ArrowArray>`, which must perform the gruesome task of applying a logical type to the arrow primitive array.


tiledb/api/src/query/read/output/arrow.rs
```diff
@@ -1,258 +1,147 @@
...
-                        expected: CellValNum::single(),
-                        found: structure.as_cell_val_num(),
-                    },
-                ))
-            }
+impl TryFrom<TypedRawReadOutput<'_>> for Arc<dyn ArrowArray> {
+    type Error = std::num::TryFromIntError;
+
+    fn try_from(value: TypedRawReadOutput<'_>) -> Result<Self, Self::Error> {
+        /*
```
rroelke, line 33: Delete commented out code

tiledb/api/src/query/read/output/arrow.rs
```diff
@@ -1,258 +1,147 @@
...
+            PT: ArrowPrimitiveTypeNative,
+            LT: LogicalType<PhysicalType = PT> + ArrowPrimitiveTypeLogical,
+            I: PrimitiveArray<<<LT as LogicalType>::PhysicalType as ArrowPrimitiveTypeNative>::ArrowPrimitiveType>,
+            O: PrimitiveArray<<LT as ArrowPrimitiveTypeLogical>::ArrowPrimitiveType>
+        {
+            unimplemented!()
+        }
+        */
+
+        fn assign_logical_type<C>(
```
rroelke, line 44: This function is what we need because the `TypedQueryBuffers` enum variants are based on the physical type, not the logical type.

When we do `QueryBufferArrowArray::from(TypedQueryBuffers)`, the primitive arrays get a field whose data type is just the physical type - e.g. all of the date/time types will be Int64.  But this can't be used later on for the RecordBatch whose schema was constructed with a date/time field - it expects the corresponding primitive type, even though the fields have the same bit width.  (as much as this was annoying here, it's probably the right call because of possible expectations of the bit representation)

I think we do need the RecordBatch to have the logical type, not the physical type - that's the way we're going to communicate with DataFusion, after all, and Tables is certainly going to want to have date/time columns.  So I believe it is important to do something here, rather than change the schema that we give to the RecordBatch.

Anyway, re-typing the primitive array was a bit tricky to figure out, but this code seems to work.

Note for the review-ee, write some of this down in a comment for this inner function.

davisp (approved):
+1 overall. Out of curiosity I went and found the Python conversion stuff and it looks like both versions mostly agree other than you've added support beyond what's in the Python bindings.

Also, it looks like they hit that same offset_size != offset_nelems issue you had as well:

https://github.com/TileDB-Inc/TileDB-Py/blob/6dfe5ce443fae1f5b16818524a8834ac7b3eb4b2/tiledb/py_arrow_io_impl.h#L696-L701

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,171 +52,563 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+                .prop_flat_map(|precision| (
+                    Just(precision),
+                    (0..precision.clamp(0, i8::MAX as u8) as i8)
+                )
+                    .prop_map(|(precision, scale)| ADT::Decimal256(
+                        precision, scale
+                    ))),
+        ];
+
+        leaf.prop_recursive(
```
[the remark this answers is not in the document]
rroelke: I figure we will do something with these eventually and so I wrote the thing down now while I was there.  But yes for the time being these just map to `DatatypeFromArrowResult::None`

tiledb/api/src/array/schema/mod.rs
```diff
@@ -432,7 +432,7 @@ impl<'ctx> Schema<'ctx> {
         Ok(c_allows_duplicates != 0)
     }
 
-    pub fn nattributes(&self) -> TileDBResult<usize> {
+    pub fn num_attributes(&self) -> TileDBResult<usize> {
```
davisp, line 435: 👀

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,169 +53,667 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+    /// There are some exceptions, such as `(Datatype::Blob, CellValNum::Var)`
+    /// mapping to `arrow::datatypes::DataType::LargeBinary`, which is always variable-length.
+    ///
+    /// ```
+    /// use tiledb::{array::CellValNum, datatype::arrow::DatatypeToArrowResult};
+    /// assert_eq!(DatatypeToArrowResult::Exact(arrow::datatypes::DataType::LargeBinary),
+    ///            tiledb::datatype::arrow::to_arrow(&tiledb::Datatype::Blob, CellValNum::Var));
+    /// ```
+    /// When the output is any kind of list, field metadata may be used to represent the exact
+    /// input datatype if the input on its own is an inexact match.
```
davisp, line 135: Am I readings this wrong or did you forget a negation in here? It feels like "field metadata may be used to represent the exact input datatype if the input is *not* an exact match". No idea what "on its own" means there.

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,169 +53,667 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+    /// ```
+    /// use arrow::datatypes::DataType as Arrow;
+    /// use tiledb::{Datatype as TileDB, array::CellValNum, datatype::arrow::DatatypeToArrowResult};
+    /// use tiledb::datatype::arrow::{to_arrow, ARROW_FIELD_METADATA_KEY_TILEDB_EXACT_MATCH};
+    /// let arrow = to_arrow(&TileDB::StringAscii, CellValNum::Var);
+    /// let DatatypeToArrowResult::Exact(Arrow::LargeList(item)) = arrow else { unreachable!() };
+    /// assert_eq!(*item.data_type(), Arrow::UInt8);
+    /// let Some(s) = item.metadata().get(ARROW_FIELD_METADATA_KEY_TILEDB_EXACT_MATCH)
+    /// else { unreachable!() };
```
davisp, line 144: Huh. TIL: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

davisp: Also the lack of formatting of doc strings makes me sad.

davisp: Sounds about Rust: It does exist, its just disabled by default and unstable. https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=format_code_in_doc_comments#format_code_in_doc_comments

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,169 +53,667 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+ * Utf8 sounds pretty good, but we can't use it because Arrow validates Utf8 and
+ * tiledb does not. So we use `LargeList(UInt8)` instead.
+ * However, in tiledb StringAscii has several special accommodations which
+ * are not granted to UInt8. We must be able to invert back to StringAscii.
+ * We can do that by storing the exact input datatype on the arrow list field metadata.
+ */
+/// `arrow::datatypes::Field` metadata key for an exact `tiledb::Datatype` variant
+/// if there is no exact mapping from `tiledb::Datatype` to `arrow::datatypes::DataType`.
+pub const ARROW_FIELD_METADATA_KEY_TILEDB_EXACT_MATCH: &str =
+    "tiledb_exact_datatype";
```
davisp, line 187: This logic is confusing me quite a bit. Are we tagging values that are an exact match and if so, why? I would expect an optional key/value that say "use this type instead" only when there's *not* an exact match. This sort of approach means we'd round trip everything we produced correctly while also transparently accepting anything we didn't produce (assuming we support it).

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,169 +53,667 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+        }
+        ADT::Binary | ADT::Utf8 | ADT::List(_) => {
+            /* offsets are 64 bits, these types use 32-bit offsets */
+            Res::None
+        }
+        ADT::BinaryView
+        | ADT::Utf8View
+        | ADT::ListView(_)
+        | ADT::LargeListView(_) => {
+            /* not something we can represent */
```
davisp, line 560: I would make this comment more specific as to the why we can't support them. Assuming we're on the same page as to the why, I'd word it as "Not something we can represent easily without re-allocating contiguous buffers and changing the schema." Or more generally "Can't be supported directly without reallocating" or whatever.

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,169 +53,667 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+            /*
+             * NB: arrow checks for valid UTF-8 but tiledb does not.
+             * This is not an exact conversion for that reason
+             * because we cannot guarantee invertibility.
+             */
+            Res::Inexact(Datatype::StringUtf8, CellValNum::Var)
+        }
+        ADT::LargeList(ref item) => {
+            if item.is_nullable() {
+                // no way to represent null values within a cell
```
davisp, line 513: What? We've got the validity buffer just like they do?

tiledb/api/src/datatype/arrow.rs
```diff
@@ -46,169 +53,667 @@ impl ArrowPrimitiveTypeNative for f64 {
...
+                Res::None
+            } else if item.data_type().primitive_width().is_none() {
+                /*
+                 * probably there are some cases we can handle,
+                 * but let's omit for now
+                 */
+                Res::None
+            } else if let Some(exact_datatype) = item
+                .metadata()
+                .get(ARROW_FIELD_METADATA_KEY_TILEDB_EXACT_MATCH)
```
davisp, line 523: I think you're doing what I'm thinking should be done, but calling this `EXACT_MATCH` is really not doing wonderful things to my brain since I consider this the precise opposite of an exact match. According to other code a TileDB Datatype::Int32 is an exact match to Arrow's Int32 type. An in-exact type would be something like "we gave arrow a Datatype::Int64 instead of a Datatype::TimeFemtosecond, so use Femtosecond instead".

We only need the extra metadata for the in-exact case so calling the extra metadata `EXACT_MATCH` strikes me as not at all correct.

Also, I'm suddenly wondering how often we'll even see this in real life? I guess joins and/or something like `insert into foo select from bar` type stuff?

davisp (changes requested):
I really don't like that `EXACT_MATCH` naming convention on the metadata key so I'm gonna ask for it to be changed. Other than that though, +1

tiledb/api/src/datatype/arrow.rs
rroelke: The validity buffer is for each cell.  For a var-size attribute the number of values per cell is variable but there is only one validity, for the whole cell.
The `item.is_nullable()` is about whether the values have a validity, not just the cell

tiledb/api/src/datatype/arrow.rs
rroelke: I added this because of some weirdness with the `StringAscii` datatype.

Let's say we wanted to do `CREATE new_table AS SELECT * FROM old_table` or something like that, preserving dimensions of the source table.  If the source table has a `StringAscii` dimension then it treats domain, extent and some other things very differently than if we used `UInt8`.

I neglected to write down how exactly this sort of thing affected the testing, and unfortunately I've forgotten... but it did, and doing this resolved it

tiledb/api/src/datatype/arrow.rs
davisp: Ahhhhhh! Gotchya. Could you reword that comment to that effect as I totes blew right past that subtlety with the comment. Just something like "TileDB supports validity of the entire cell, but not for values within the cell so we can't support this". I'm guessing my brain read "TileDB has no way to represent validity" and then clicked off to "What? Yes it do!" before finishing the "within a cell" part.

tiledb/api/src/datatype/arrow.rs
davisp: Sure. The TileDB <=> Arrow round tripping is lossy due to type mismatches. Adding type hints to aid in round trip fidelity makes perfect sense. I just think the naming is wrong since we only need the hint in the absence of an exact match.

davisp (approved):
+1'ed

PR #89, opened 2024-05-08
Remove restrictions on datatypes for proptest strategies

rroelke:
To date, we were actually only using a few different datatypes for the proptest strategies.  And we were missing all cell val nums other than 1.

This pull request updates our strategies to use all datatypes, and consider other cell val nums as well.

There are a few issues which are revealed by those changes:
- restrictions on RLE and Dictionary filters in pipelines (fixed in this PR)
- multiple-value fill values for attributes (fixed in this PR)
- dimension domain/extent sometimes silently ignored (deferred)

### DIFF
--- a/tiledb/api/src/datatype/strategy.rs
+++ b/tiledb/api/src/datatype/strategy.rs
@@ -51,7 +51,7 @@ fn prop_datatype() -> impl Strategy<Value = Datatype> {
     ]
 }
 
-const DENSE_DIMENSION_DATATYPES: [Datatype; 31] = [
+const DENSE_DIMENSION_DATATYPES: [Datatype; 30] = [
     Datatype::Int8,
     Datatype::Int16,
     Datatype::Int32,
@@ -82,10 +82,9 @@ const DENSE_DIMENSION_DATATYPES: [Datatype; 31] = [
     Datatype::TimePicosecond,
     Datatype::TimeFemtosecond,
     Datatype::TimeAttosecond,
-    Datatype::Boolean,
 ];
 
-const SPARSE_DIMENSION_DATATYPES: [Datatype; 34] = [
+const SPARSE_DIMENSION_DATATYPES: [Datatype; 33] = [
     Datatype::Int8,
     Datatype::Int16,
     Datatype::Int32,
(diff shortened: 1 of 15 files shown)


tiledb/api/src/datatype/strategy.rs
```diff
@@ -64,24 +62,76 @@ pub fn prop_datatype_implemented() -> impl Strategy<Value = Datatype> {
...
+        Just(Datatype::DateTimeAttosecond),
+        Just(Datatype::TimeHour),
+        Just(Datatype::TimeMinute),
+        Just(Datatype::TimeSecond),
+        Just(Datatype::TimeMillisecond),
+        Just(Datatype::TimeMicrosecond),
+        Just(Datatype::TimeNanosecond),
+        Just(Datatype::TimePicosecond),
+        Just(Datatype::TimeFemtosecond),
+        Just(Datatype::TimeAttosecond),
```
davisp, line 86: You're missing `Datatype::Bool` here. Its weird but that's core's definition.

tiledb/api/src/datatype/strategy.rs
```diff
@@ -64,24 +62,76 @@ pub fn prop_datatype_implemented() -> impl Strategy<Value = Datatype> {
...
+        Just(Datatype::StringAscii),
```
davisp, line 125: Also bool as above.

davisp (approved):
Jus two minor notes. Otherwise +1

tiledb/api/src/datatype/strategy.rs
rroelke: This was a good note to add a check that everything outside of this set cannot be used for a dense dimension.

rroelke: Updated the test in a way that you will hopefully like - we construct a schema and check that it succeeds only if we expected it to, and if it failed we make sure it's because of a datatype error.

davisp: I do indeed like that! The only reason I hadn't discovered that is_byte_type disparity is because I hadn't gotten to testing dimensions directly yet.

PR #91, opened 2024-05-08
Port array metadata functions.

abigalekim:


rroelke:
This looks quite reasonable, it's nice that it layered onto the previous work with groups so well.

There does need to be tests here for sure, I'd suggest also covering the mode requirements that are covered in the C header doc strings.  Apparently some of these functions can only be called in read mode? Weird.

Please add doc strings too, we definitely want to improve our efforts with those, otherwise it's going to be a massive pain to go back and do them all as we approach a release.

davisp (commented):
I'll leave final approval to @rroelke since he's done other reviews. But from my skimming I've not seen any issues and would be +1.

rroelke (approved):
Looks good, except do consider adding some docstrings prior to merging, otherwise it's going to be a pain to go and do them all prior to... well I'm not sure what exactly but some kind of point of publishing

PR #92, opened 2024-05-09
Make DimensionData domain and extent optional

davisp:
This updates the DimensionData type to have an optional domain and extent to cover more combinations of cases.


tiledb/api/src/array/dimension/mod.rs
```diff
@@ -207,62 +209,63 @@ impl<'ctx> Builder<'ctx> {
         })
     }
 
-    /// N.B., This API will eventually be removed when we fixup the current
-    /// implementation of `new` to take a `DimensionConstraint` argument instead
-    /// of the serde values for range and extent.
-    pub fn new_string(
+    pub fn new_optional(
```
davisp, line 212: I'm not a super huge fan of this name, but I couldn't come up with anything better.

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -207,62 +209,63 @@ impl<'ctx> Builder<'ctx> {
...
-            )
-        })?;
-        Ok(Builder {
-            dim: Dimension {
-                context,
-                raw: RawDimension::Owned(c_dimension),
-            },
-        })
-    }
+        let mut domain_data: Vec<u8> = Vec::new();
```
rroelke, line 223: Ok I think I follow why you need `Vec` here - because of the enum variants not using a slice, it might not be guaranteed that the values are contiguous, this fixes things in memory the way you want.

Regarding the endian-ness... does tiledb use the native byte ordering?
If it does, then converting may be incorrect.
If it does not, then we probably have a lot more work to do to get things to line up.

For the extent, seems like you could do `value as *const DT as *const std::ffi::c_void` and avoid a vec there.

tiledb/api/src/array/dimension/mod.rs
rroelke: There's no guidance on which to use, `new` or `new_optional`. Perhaps writing it would inspire the right name.

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -207,62 +209,63 @@ impl<'ctx> Builder<'ctx> {
...
 
-    /// N.B., This API will eventually be removed when we fixup the current
-    /// implementation of `new` to take a `DimensionConstraint` argument instead
-    /// of the serde values for range and extent.
-    pub fn new_string(
+    pub fn new_optional(
         context: &'ctx Context,
         name: &str,
         datatype: Datatype,
+        domain: Option<SingleValueRange>,
```
rroelke, line 216: I don't think this is sound without a safeguard that `domain` and `extent` are the same physical type.

tiledb/api/src/array/dimension/mod.rs
rroelke: I think I still like my suggestion to have one function the most with something like
```
pub enum DimensionDatatype {
    UInt8 { domain: std::ops::Range<u8>, extent: Option<u8> },
    UInt64 { domain, std::ops::Range<u64>, extent: Option<u64> },
    ...,
    StringAscii,
    StringUtf8
}
```

tiledb/api/src/array/dimension/mod.rs
davisp: AFAIK, TileDB just assumes the entire world is little endian. I don't think there's anything in core that tries to handle running on big endian architectures. Although I have found a few places where we store data on disk as big endian for reasons I have yet to fathom. (If memory serves, one of the filter pipeline option values is stored big endian).

For the Vec, that's pretty much correct. I needed contiguous bytes in a type safe allocation that outlives the call to `tiledb_dimension_alloc` and that was the least worse way I could think to do it. Though I'll likely revisit this after I finish figuring out that bug that proptests found yesterday.

tiledb/api/src/array/dimension/mod.rs
davisp: I'm mulling over redoing the whole Builder/Constructor pattern. One thought I had to clean this up quite a bit would be to switch this Builder to be more like the Enumeration version where the only FFI calls happen in `build()` and everything else is just setting values on a DimensionData instance. Although, even then I'm not sure how to make new() work with and without a Range that isn't super misleading to users. I'll keep thinking on it.

davisp: Yeah, I might go that way. I started to go that route originally but it felt weird so I didn't follow through on it. Though now that I've got the current shape of things I'll give it another whirl to see if clears up the issues with `new`.

tiledb/api/src/array/dimension/mod.rs
davisp: Its absolutely not. Though I swear I had type checks in here for this specific reason. I was probably distracted by something shiny and put them in the wrong place or something.

tiledb/api/src/array/dimension/mod.rs
davisp: Any reason you went with std::ops::Range over `[T; 2]`? The array approach means that Vec allocation goes away which would be nice.

rroelke: You're right, `[T; 2]` should be better for that.  It is nice that range carries the semantic meaning though.

I guess you could construct `[T; 2]` trivially from a range, so you wouldn't need Vec even with the range.

I don't really have a preference

rroelke: > I'm mulling over redoing the whole Builder/Constructor pattern. One thought I had to clean this up quite a bit would be to switch this Builder to be more like the Enumeration version where the only FFI calls happen in build() and everything else is just setting values on a DimensionData instance

I'm not sure how well that solves this particular problem but I have had this thought that this is something we could apply more broadly.  It would be kind of nice to get the `TileDBResult` out of so many of the setter functions

davisp: Yeah, I think the DimensionDatatype will likely end up being cleaner for the API rather than whatever I was thinking. I'm gonna go start my caffeination protocols and will try that approach.

Totes agree that those Result's going away would be nice. I've also noticed that the using the `impl Factor for $ThingData` is quite lovely for a more programatic API interaction.

tiledb/api/examples/quickstart_sparse_string.rs
```diff
@@ -53,14 +52,16 @@ fn create_array(ctx: &Context) -> TileDBResult<()> {
                 DimensionData {
                     name: "rows".to_owned(),
                     datatype: Datatype::StringAscii,
-                    ..Default::default()
+                    constraints: ().into(),
```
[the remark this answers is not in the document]
rroelke: Ok, I don't really think `Default` is the right choice actually, but I feel like rquiring the user to write `DimensionConstraint::StringAscii` is not a big ask, and the readability is going to be much better than `().into()`

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -164,101 +163,219 @@ impl<'c1, 'c2> PartialEq<Dimension<'c2>> for Dimension<'c1> {
...
+    }
+
+    pub fn domain_ptr(&self) -> *const std::ffi::c_void {
+        dimension_constraints_go!(
+            self,
+            DT,
+            range,
+            _extent,
+            range.as_ptr() as *const DT as *const std::ffi::c_void,
+            std::ptr::null()
```
rroelke, line 327: Can these return `Option<*const std::ffi::c_void>`?  Feels better (to me) than checking for null pointer.

`std::ptr::NonNull` looks interesting too but it assumes `*mut` which isn't appropriate here.

Also these should probably be `pub(crate)` instead of `pub`, IMO we shouldn't leak any FFI types.

tiledb/api/src/array/dimension/strategy.rs
```diff
@@ -72,24 +72,27 @@ where
         )
     })
     .prop_flat_map(move |(lower_bound, upper_bound)| {
-        let extent_limit = {
+        let (extent_limit, would_overflow) = {
```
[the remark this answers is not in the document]
davisp: This is definitely a gnarly bit of code and the only confidence I have in it being correct is due to the proptests not breaking anymore. And even then I'm merely "pretty sure" its correct.

However, it didn't seem too bad with the knowledge that eventually we're going to be able to remove that extra check once core fixes that bug its avoiding.

tiledb/api/src/datatype/strategy.rs
```diff
@@ -210,19 +208,23 @@ mod tests {
...
+        match r {
+            Err(Error::LibTileDB(s)) => {
+                assert!(
+                    s.contains("not a valid Dimension Datatype")
+                        || s.contains("do not support dimension datatype"),
+                    "Expected dimension datatype error, received: {}",
+                    s
+                );
+            }
+            Err(Error::Datatype(reason)) => {
```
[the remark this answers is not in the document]
davisp: You've confused me with this one. All of the usable datatypes don't return an error so they're not a concern here? We're only checking that the failures are failing because of an invalid type error and not some other random issue. What I *think* you mean is "If the error isn't coming from core, we might miss when they enable a new dimension type" which is reasonable.

I'll update the test to avoid passing all of the string types through with the `DimensionConstraint::StringAscii` which rejects all of those other types.

tiledb/api/examples/quickstart_sparse_string.rs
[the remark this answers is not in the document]
davisp: Totes fair on this one as this felt less than super awesome when I was writing it but I couldn't make a decision either direction on what to do. I'll totally rip out that ().into implementation as I agree its even less great looking at it again with fresh eyes, and it wasn't all that great when I wrote it.

I'm not a super big fan of having. to write out `DimensionConstraint::StringAscii` on the line after `Datatype::StringAscii` since it really seems like the constraint should be inferable there. But its at least better than what I've got currently at least.

tiledb/api/src/datatype/strategy.rs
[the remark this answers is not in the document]
davisp: Fixed. Error checking is now 100% untouched so that its core flagging acceptability on the datatype or not.

tiledb/api/src/array/dimension/mod.rs
rroelke: Good point that this is not checked but passed straight to the FFI call.  Option would just be a headache then.

tiledb/api/examples/quickstart_sparse_string.rs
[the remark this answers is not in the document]
rroelke: `DimensionConstraint::None` or `DimensionConstraint::Unconstrained`?

I see your point and I kind of agree, especially since the `StringAscii` might well be relaxed in future versions such that other types can also be variable in a dense dimension

tiledb/api/src/datatype/strategy.rs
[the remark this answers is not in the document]
rroelke: Er, yes, you interpret what I mean correctly. Thank you

PR #93, opened 2024-05-13
Remove QueryCondition struct, user just provides enum

rroelke:
This is a minor tweak to the query condition API which simplifies the experience for users a bit.  The `struct QueryCondition` is only ever build and then immediately passed to the query builder API.  Any manipulations will be done by the enum.  Given that, why should the user need to build a `struct QueryCondition` at all? Let's just have the query do it.  That's what this PR implements.


davisp (approved):
+1

PR #94, opened 2024-05-13
FillData::data is not json but instead metadata::Value

rroelke:
When a proptest fails with schema data, we would really like to be able to copy/paste the `Debug` output into a unit test so we can step through the execution of the minimum failing input with ease.

At present there are a number of alterations required in order to copy/paste.  Attribute fill data, for example, looks like:
```
fill: Some(
                    FillData {
                        data: Array [
                            Number(52),
                            Number(78),
                            Number(56),
                            Number(57),
                            Number(-6),
                            Number(-81),
                            Number(-89),
                            Number(23),
                            Number(108),
                            Number(36),
                        ],
                        nullability: Some(
                            false,
                        ),
                    },
                )
```
This can probably be copy/pasteable by doing `use serde_json::Value::*` or something along those lines.  But we'd probably rather see something like `data: UInt64([52, 78, 56, 57, -6, -81, -89, 23, 108, 36])`.

This change also is a step towards removing `serde_json` as a dependency for our `api` crate, if we determine we would prefer that.

### DIFF
--- a/tiledb/api/src/array/attribute/mod.rs
+++ b/tiledb/api/src/array/attribute/mod.rs
@@ -582,7 +582,7 @@ impl<'a> FromFillValue<'a> for String {
 /// Encapsulation of data needed to construct an Attribute's fill value
 #[derive(Clone, Debug, Deserialize, OptionSubset, PartialEq, Serialize)]
 pub struct FillData {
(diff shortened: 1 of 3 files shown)


davisp (approved):
+1

PR #95, opened 2024-05-14
Prerequisite changes for generating an arbitrary schema and write query

rroelke:
The changes needed to generate an arbitrary schema and write query for that schema are quite expansive.  This pull request consists of various prerequisite changes that I have accumulated while identifying problems and fixing bugs in the goal strategies.  My intent is that by reviewing these separately we can have a more focused discussion of these prerequisites and then later a more focused discussion of the changes to the test strategies.


davisp (approved):
+1

PR #96, opened 2024-05-15
Properly use cell_val_num in DataProvider impls

rroelke:
This argument was tacked on in prior work, but as we turn more attention to the write side of queries it deserves some proper attention.

Different inputs may behave differently with different `cell_val_num`, or may even error out.  `DataProvider::as_tiledb_input` should return `TileDBResult`, and our existing impls should properly use `cell_val_num`.


davisp (approved):
+1

PR #97, opened 2024-05-17
Implement query input traits for Arrow data

rroelke:
Previous work turned query read output into arrow data.  This pull request flips that, turning arrow data into query input.

This involves several implementations of `DataProvider` for arrow array types, as well as introducing a similar `TypedDataProvider` which we need to do `impl TypedDataProvider for dyn ArrowArray`.  `DataProvider` has an object create `QueryBuffers` and `TypedDataProvider` has an object return `TypedQueryBuffers`.  These implementations are tested by taking something resembling tiledb query output, turning it into arrow, and then turning the arrow back into the tiledb format.

We also add a new trait `RecordProvider` which we implement for `RecordBatch`, adding a set of named fields to a write query.  This is not tested yet and probably won't be while the write test strategy is still pending.

### DIFF
--- a/tiledb/api/src/query/read/output/strategy.rs
+++ b/tiledb/api/src/query/read/output/strategy.rs
@@ -221,11 +221,11 @@ where
                             .into_iter()
                             .take(ncells + 1)
                             .collect::<Vec<u64>>();
-                        if let Some(first) = offsets.first_mut() {
-                            *first = 0u64;
-                        }
                         if let Some(last) = offsets.last_mut() {
                             *last = nvalues as u64;
+                        }
+                        if let Some(first) = offsets.first_mut() {
+                            *first = 0u64;
                         }
                         offsets.sort();
 
--- a/tiledb/api/src/query/write/input/arrow.rs
+++ b/tiledb/api/src/query/write/input/arrow.rs
@@ -2,10 +2,11 @@ use std::rc::Rc;
 use std::sync::Arc;
 
 use arrow::array::{
-    Array as ArrowArray, FixedSizeBinaryArray, FixedSizeListArray,
+    Array as ArrowArray, AsArray, FixedSizeBinaryArray, FixedSizeListArray,
     GenericListArray, LargeBinaryArray, LargeStringArray, PrimitiveArray,
     RecordBatch,
 };
+use arrow::buffer::OffsetBuffer;
 use arrow::datatypes::{ArrowPrimitiveType, Field};
 
 use crate::array::{CellValNum, Schema};
@@ -18,6 +19,25 @@ use crate::query::write::input::{
     DataProvider, RecordProvider, TypedDataProvider,
 };
 use crate::Result as TileDBResult;
+
+fn cell_structure<'data>(
+    offsets: &'data OffsetBuffer<i64>,
+    cell_val_num: CellValNum,
+) -> TileDBResult<CellStructure<'data>> {
+    match cell_val_num {
+        CellValNum::Fixed(nz) => {
+            let expect_len = nz.get() as i64;
+            for window in offsets.windows(2) {
+                if window[1] - window[0] != expect_len {
+                    /* TODO: error */
+                    unimplemented!()
+                }
+            }
+            Ok(CellStructure::Fixed(nz))
+        }
+        CellValNum::Var => Ok(CellStructure::Var(Buffer::<u64>::from(offsets))),
+    }
+}
 
 fn validity_buffer<A>(
     array: &A,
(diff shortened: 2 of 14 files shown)


rroelke:
There are still lots of `todo!()` and `unimplemented!()` for the error handling, these must be filled in prior to merging.

tiledb/api/src/query/read/output/strategy.rs
```diff
@@ -221,12 +221,12 @@ where
...
                             .collect::<Vec<u64>>();
-                        if let Some(first) = offsets.first_mut() {
-                            *first = 0u64;
-                        }
                         if let Some(last) = offsets.last_mut() {
                             *last = nvalues as u64;
                         }
+                        if let Some(first) = offsets.first_mut() {
+                            *first = 0u64;
+                        }
```
davisp, line 229: This is just out of curiosity, but am I missing a reason for this or is it just a "temporary thing that got refactored out" change?

tiledb/api/src/query/write/input/arrow.rs
```diff
@@ -0,0 +1,753 @@
...
+        CellValNum::Fixed(nz) => Ok(CellStructure::Fixed(nz)),
+        CellValNum::Var => {
+            let offsets = Buffer::Owned(
+                std::iter::repeat(fixed_len as usize)
+                    .take(ncells)
+                    .enumerate()
+                    .map(|(i, len)| (i * len) as u64)
+                    .collect::<Vec<u64>>()
+                    .into_boxed_slice(),
+            );
```
davisp, line 78: Not asking for a change here, but did you try looking at the assembly or one of the IR outputs to get an idea on the efficiency for this? I'm curious if going the other route wouldn't be faster. Something like:

```rust
let data = Vec::<u64>::new();
data.iter_mut().enumerate().for_each(|idx, val| *val = idx * len);
```

I'm mostly just wondering if one or the other would get pushed down to a single allocation with a single auto vectorized loop for initialization or not.

tiledb/api/src/query/write/input/arrow.rs
```diff
@@ -0,0 +1,753 @@
...
+            | ADT::Union(_, _)
+            | ADT::Dictionary(_, _)
+            | ADT::Decimal128(_, _)
+            | ADT::Decimal256(_, _)
+            | ADT::Map(_, _)
+            | ADT::RunEndEncoded(_, _) => {
+                Err(Error::InvalidArgument(anyhow!(format!(
+                    "Unsupported Arrow datatype as query input: {}",
+                    self.data_type()
+                ))))
```
davisp, line 457: Was there a particular reason for enumerating all of these instead of having a `_ =>` default arm? It doesn't bother me either direction, I just can't come up with a reason other than just trying to be specific in case this list changes in the future.

davisp (approved):
+1 I only had a couple questions for personal insight when I wasn't 100% on a change. Everything else looks fine.

tiledb/api/src/query/read/output/strategy.rs
rroelke: This one actually was functional!

(side note: I think it would be wise to also have the write side do the extra offsets, for consistency, plus we can likely get rid of this sort of thing)

Because the write side doesn't have the extra offset, if you have one cell, then you have one offset.  If you have one offset, then `first_mut()` and `last_mut()` return the same memory location.  The first offset is always supposed to be zero, but it wouldn't always be so without swapping the order!

tiledb/api/src/query/write/input/arrow.rs
rroelke: I wrote it down this way before filling in the error.  I wanted to make sure I didn't miss anything.

Now that I haven't, we could change it, but my inclination is to leave it as is just to be extra sure we catch everything if arrow introduces a new type.

tiledb/api/src/query/write/input/arrow.rs
rroelke: Haven't done that yet.  Seems like a good exercise.

davisp: I don't remember what I was doing exactly, but it was something like making sure that something I did was either a memcpy or memset instead of iterating over an entire vector. If memory serves there's a blog post or SO post that says what compiler output level to look at to see if its inserting a syscall or not. I'm not entirely sure how to check for auto vectorization.

rroelke: The version you offer is fewer instructions and looks simpler so I'll pivot to that

davisp: Cool beans!

tiledb/api/src/query/read/output/strategy.rs
rroelke: The number of values of cell `i` is determined by `offsets[i + 1] - offsets[i]` and are element-based.  And then `offsets[i]` is the index of the first value for cell `i`.

```
                proptest::collection::vec(
                    0u64..=(nvalues as u64),
                    0..=max_offsets,
                ),
```
This part of the strategy picks a random set of indices into the values.

```
                        offsets = offsets
                            .into_iter()
                            .take(ncells + 1)
                            .collect::<Vec<u64>>();
                        if let Some(last) = offsets.last_mut() {
                            *last = nvalues as u64;
                        }
                        if let Some(first) = offsets.first_mut() {
                            *first = 0u64;
                        }
                        offsets.sort();
```
This part of the strategy makes sure that the set of random indices conforms to the other required properties, which is that (1) the offsets begin at zero; (2) the offsets end with the number of values; and (3) the offsets are sorted.

If we have 10 values then maybe the initial part gives us `[4, 6, 3, 7, 1, 1, 3]`, which would turn into `[0, 6, 3, 7, 1, 1, 10]` before sorting, and then `[0, 1, 1, 3, 6, 7, 10]` after sorting.

davisp: I told you my brain wasn't working last night. I was totally missing the `offsets.sort()` step. Nice!

PR #99, opened 2024-05-21
feat: [sc-47897] [rs] Array nonempty_domain functions

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/47897

I haven't added any testing in this PR yet, I would like to test this with the array write proptest in #98 .  This will be tested by whichever lands first.

### DIFF
--- a/tiledb/api/src/array/mod.rs
+++ b/tiledb/api/src/array/mod.rs
@@ -1,17 +1,21 @@
 use std::convert::TryFrom;
 use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+use std::num::NonZeroU32;
 use std::ops::Deref;
 
+use anyhow::anyhow;
 use serde::{Deserialize, Serialize};
 use util::option::OptionSubset;
 
 use crate::array::schema::RawSchema;
 use crate::context::{CApiInterface, Context, ContextBound};
-use crate::error::ModeErrorKind;
+use crate::datatype::{LogicalType, PhysicalType};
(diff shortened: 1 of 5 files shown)


tiledb/api/src/array/mod.rs
```diff
@@ -367,6 +371,291 @@ impl Array {
...
+        let (datatype, cell_val_num) = {
+            let dim = self.schema()?.domain()?.dimension(key.clone())?;
+            (dim.datatype()?, dim.cell_val_num()?)
+        };
+        self.dimension_nonempty_domain_impl(key, datatype, cell_val_num)
+    }
+
+    /// Returns the non-empty domain of all dimensions from this array, if any.
+    /// The "non-empty domain" is the set of coordinates across all dimensions
+    /// which have had data written into them.
```
davisp, line 648: This comment and the one above seem a bit oddly worded to me. I wouldn't call it the "set" of coordinates, though two does technically count there.

How about something along the lines of "The domain of a dimension is set at schema creation time. The non-empty domain is the minimum and maximum coordinate value that has been written to disk so far."

davisp (approved):
+1

tiledb/api/src/array/mod.rs
rroelke: I took this suggestion and added a bit more to it.

This is why I think it's important we do more of the docs as we go - they will each get more attention than they would with one big PR before publishing time :)

PR #100, opened 2024-05-23
Update CI to use nightly builds of core

davisp:
Yep.


tiledb/api/src/filter/strategy.rs
```diff
@@ -83,53 +83,51 @@ fn prop_bitwidthreduction() -> impl Strategy<Value = FilterData> {
...
-                    },
-                )
-                .prop_map(|dt| CompressionType::DoubleDelta {
-                    reinterpret_datatype: Some(dt),
-                });
-
-            return vec![delta.boxed(), double_delta.boxed()];
+    fn validate_type(
+        in_dtype: &Option<Datatype>,
+        out_dtype: &Datatype,
```
rroelke, line 88: I stared at this a lot longer than I probably would have if it were named `reinterpret_datatype`.
The reinterpret datatype isn't the same as the output type, the output type will soon be blob (https://github.com/TileDB-Inc/TileDB/pull/4988/files#diff-f46f30278b5b0f9992c18dab02d507dc5f489bc304d82a6404c692da761ddaf3)

rroelke (approved):
A suggested tweak for an argument name but otherwise - let it roll

rroelke:
Actually hold up on that approval, I should have checked out the branch first.
It is not building for me
```
  --- stderr
  thread 'main' panicked at tiledb/sys/build.rs:5:10:
  Missing tiledb dependency.: `PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --variable=libdir tiledb` did not exit successfully: exit status: 1
  --- stderr
  Package tiledb was not found in the pkg-config search path.
  Perhaps you should add the directory containing `tiledb.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'tiledb' found
```

hopefully I am capable of figuring this out myself, we will soon see

rroelke:
This is great, it fixes some issues that `cargo test --doc --all-features` was having for me where there would be a runtime error finding libtiledb for the arrow doc tests.

ihnorton (approved):

PR #102, opened 2024-05-28
Replace TempDir with TestArrayUri

davisp:
This removes our usage of the TempDir for tests and instead uses a new TestArrayUri trait. Eventually this will allow us to dynamically change whether tests run against a local file system or a remote cloud service instance.

This does not yet support actually running against the cloud service as that will take more infra setup to get going. However, it should be as simple as creating a new TestCloudUri struct which then implements the TestArrayUri trait. Then the `test_util::get_uri_generator` can return either trait implementation at runtime based on an environment variable or whatever is easiest in CI.

### DIFF
--- a/tiledb/api/src/array/fragment_info.rs
+++ b/tiledb/api/src/array/fragment_info.rs
@@ -744,18 +744,17 @@ impl Builder {
 #[cfg(test)]
 pub mod tests {
     use super::*;
-    use tempfile::TempDir;
-
     use crate::array::*;
     use crate::config::Config;
     use crate::query::{QueryBuilder, WriteBuilder};
+    use crate::test_util::{self, TestArrayUri};
     use crate::Datatype;
 
     #[test]
     fn test_set_config() -> TileDBResult<()> {
         let ctx = Context::new().unwrap();
-        let tmp_dir = TempDir::new().unwrap();
-        let array_uri = create_dense_array(&ctx, &tmp_dir).unwrap();
+        let test_uri = test_util::get_uri_generator()?;
+        let array_uri = create_dense_array(&ctx, &test_uri)?;
 
         let config = Config::new()?;
         let frag_infos =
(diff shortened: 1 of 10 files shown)


tiledb/api/src/array/fragment_info.rs
```diff
@@ -744,18 +744,17 @@ impl Builder {
 #[cfg(test)]
 pub mod tests {
     use super::*;
-    use tempfile::TempDir;
-
     use crate::array::*;
     use crate::config::Config;
     use crate::query::{QueryBuilder, WriteBuilder};
+    use crate::test_util::{self, TestArrayUri};
```
rroelke, line 750: Huh, what does `self` mean here?

davisp: Its the equivalent to `use crate::test_util` so that you can write `test_util::get_uri_generator()`.

tiledb/api/src/test_util.rs
```diff
@@ -0,0 +1,81 @@
...
+        })
+    }
+}
+
+impl TestArrayUri for TestDirectory {
+    fn base_dir(&self) -> TileDBResult<String> {
+        self.base_dir()
+    }
+
+    fn with_path(&self, path: &str) -> TileDBResult<String> {
```
rroelke, line 70: This resolves correctly?
It must, since you have a passing CI.  I'm surprised though.  Does it break if you `use` the trait?

tiledb/api/src/test_util.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.with_path(path)
```
rroelke, line 71: Do you ever envision a different implementation? This could be a provided method

rroelke (approved):

tiledb/api/src/array/fragment_info.rs
rroelke: Neat!

tiledb/api/src/test_util.rs
davisp: Yeah, I waffled on that but had written the TestDirectory implementation before defining the trait. I'll go back and make it a provided method.

tiledb/api/src/test_util.rs
davisp: Not sure what resolution you're asking about here? The `self.with_path(path)` call and how its picked the TestDirectory implementation vs the trait-method-which-becomes-recursive issue? I contemplated if I'd need to cast things there but just went with the obvious first and it worked so I stopped thinking about it.

I *think* the "cast required to pick an implementation" only kicks in when you have multiple traits with duplicate names but I'd have to go read the spec closely to get to the root of it.

tiledb/api/src/test_util.rs
davisp: Turns out the VFS tests (which can't run against the REST api) were using this so I had to keep the implementation on `TestDirectory`. Though its now provided so the `impl TestArrayUri for TestDirectory` doesn't have to specify it at least.

PR #103, opened 2024-05-28
Proptest for checking correctness of a single sparse write

rroelke:
This is the follow-up to #98 , extending the proptest to also work correctly for sparse writes.

Things this pull request had to deal with:
- filters on schema coordinates and dimensions because those can be Float types now
- generating unique values for dimensions when required

There are a few notable tools used to accomplish these goals.

As we've been building out the API we've had more and more reasons to separate handling of floating-point types from integral types for the various enums that can hold both. This pull request encounters yet another, and so we add the `crate::datatype::Integral` trait as a helper for writing generic code over u8, u16, u32, etc.

Another tool added here is an analog of `Iterator::map` for the query callback adapters. We add `map_intermediate`, `map_final`, and `map` functions for transforming the result of a callback into something else.

### DIFF
--- a/tiledb/api/src/array/schema/mod.rs
+++ b/tiledb/api/src/array/schema/mod.rs
@@ -390,18 +390,14 @@ impl Schema {
         Ok(Schema::new(context, RawSchema::Owned(c_schema)))
     }
 
-    pub fn version(&self) -> TileDBResult<i64> {
+    pub fn version(&self) -> TileDBResult<u32> {
         let c_schema = self.capi();
-        let mut c_version: std::os::raw::c_int = out_ptr!();
+        let mut c_version: u32 = out_ptr!();
         self.capi_call(|ctx| unsafe {
-            ffi::tiledb_array_schema_get_allows_dups(
-                ctx,
-                c_schema,
-                &mut c_version,
-            )
+            ffi::tiledb_array_schema_get_version(ctx, c_schema, &mut c_version)
         })?;
 
-        Ok(c_version as i64)
+        Ok(c_version)
     }
 
     pub fn array_type(&self) -> TileDBResult<ArrayType> {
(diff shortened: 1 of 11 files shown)


tiledb/api/src/array/schema/mod.rs
```diff
@@ -374,18 +390,14 @@ impl Schema {
         Ok(Schema::new(context, RawSchema::Owned(c_schema)))
     }
 
-    pub fn version(&self) -> TileDBResult<i64> {
+    pub fn version(&self) -> TileDBResult<u32> {
         let c_schema = self.capi();
-        let mut c_version: std::os::raw::c_int = out_ptr!();
+        let mut c_version: u32 = out_ptr!();
         self.capi_call(|ctx| unsafe {
-            ffi::tiledb_array_schema_get_allows_dups(
```
davisp, line 381: Whoops

tiledb/api/src/array/schema/mod.rs
```diff
@@ -984,7 +1009,12 @@ mod tests {
         .unwrap();
 
         let s: Schema = b.build().unwrap();
-        assert_eq!(0, s.version().unwrap());
+
+        let schema_version = s.version().unwrap();
+
+        // it's a little awkward to use a constant here but this does
+        // not appear to be tied to the library version
```
davisp, line 1016: Correct. The schema version is actually the underlying data format which is incremented anytime we change how something is stored on disk. You can find the mapping of storage format versions to library versions here:

https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/FORMAT_SPEC.md#history

tiledb/api/src/array/schema/mod.rs
```diff
@@ -1566,4 +1596,73 @@ mod tests {
             assert_ne!(base, cmp);
         }
     }
+
+    /// Test what the default values filled in for `None` with schema data are.
+    /// Mostly because if we write code which does need the default, we're expecting
+    /// to match core and need to be notified if something changes or we did something
+    /// wrong.
```
davisp, line 1603: Excellent idea!

davisp (approved):
+1

PR #104, opened 2024-05-28
Fix FixedDataIterator and VarDataIterator soundness

rroelke:
We have tests `test_fixed_data_iterator_lifetime` and `test_var_data_iterator_lifetime` which demonstrate unsoundness of an `unsafe` block inside `FixedDataIterator` and `VarDataIterator` respectively.  This unsoundness arises when there is a transfer of ownership of the values data into the iterator, potentially resulting in a situation where an item yielded lives longer than the iterator does.

This pull request fixes that unsoundness by adding a type-level proof that the data passed to the iterator is *borrowed*, not owned.  The user cannot construct an iterator which satisfies the conditions for unsoundness.

### DIFF
--- a/tiledb/api/src/query/buffer/mod.rs
+++ b/tiledb/api/src/query/buffer/mod.rs
@@ -457,12 +457,13 @@ macro_rules! query_buffers_proof_impls {
         )+
     }
 }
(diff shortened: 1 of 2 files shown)


davisp (approved):
+1

PR #107, opened 2024-05-30
feat: [sc-48511] Fix sparse array write strategy "too many local rejects"

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/48511

> Generating a write input for a sparse array may require generating unique coordinates.
>
> The current strategy generates coordinates which are unique for each dimension individually. If a single dimension is constrained to the domain 0.. 10, for example, then the strategy cannot possibly produce more than 10 records of input, no matter how many other dimensions there are.
> 
> This means that the test generator will never cover the whole space of a sparse dimension.
> 
> The implementation of choosing the unique values within a dimension also is not very smart. It uses proptest::collection::btree_set, which generates a bunch of values and then throws the whole thing away to try again if it is below the requested number of records. The probability of rejection increases a lot as the dimension gets more and more constrained, and we have seen in the occasional run that proptest eventually gives up if it cannot satisfy the minimum number of records.
>
> For all of the above we must change our strategy to choose unique values across all dimensions instead.

This pull request fixes the issue by modifying the `Cells` strategy to first generate the maximum number of records for the dimensions, then deduplicate the coordinates.  Then we generate the rest of the attributes to write using the new number of records.

### DIFF
--- a/tiledb/api/src/datatype/physical.rs
+++ b/tiledb/api/src/datatype/physical.rs
@@ -104,6 +104,13 @@ where
     }
 }
 
+/// Trait for hashing based on value bits.
+/// This exists to work around float types, which do not implement `Hash`.
+/// That makes generic programming on all physical types more challenging.
+///
+/// Types implementing `BitsHash` can be hashed by an instance of `Hasher`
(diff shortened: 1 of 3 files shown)


davisp (approved):
+1

PR #109, opened 2024-05-31
feat: [sc-48380] [rs] Proptest which checks correctness of a sequence of writes to a sparse array

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/48380

For each write in the sequence, we check the correctness of:
1) the non-empty domain of the most recent fragment
2) the non-empty domain of the whole array
3) the contents read back from the most recent fragment
4) the contents read back from the whole array

Items 2 and 4 require opening the array over a particular timestamp range, so we add the `ArrayOpener` class to configure the attributes of an array before it is open.

Otherwise, this builds nicely on top of the test cells de-duplication logic added in #107 .


tiledb/api/src/array/mod.rs
```diff
@@ -661,21 +658,111 @@ impl Array {
...
+            array: Array {
+                context: context.clone(),
+                uri: uri.as_ref().to_owned(),
+                raw: RawArray::Owned(array_raw),
+            },
+            mode: Some(mode),
+        })
+    }
+
+    /// Configures the start timestamp for an open array.
```
davisp, line 724: I think this phrasing is slightly wrong? These can't be set after an array is open (excluding the re-open semantics). So it should say something like "for an array about to be opened" or similar?

tiledb/api/src/query/write/strategy.rs
```diff
@@ -976,19 +1162,63 @@ mod tests {
...
+            };
+
+            /*
+             * Then re-open the array to read back what we just wrote
+             * into the most recent fragment only
+             */
             {
+                array = array
+                    .refresh()
+                    .start_timestamp(timestamp_min)
```
davisp, line 1215: Did I miss a sleep in here or a place where we set these timestamps for writes? There's an issue that folks run into (mostly while testing) where its possible to write two fragments within the same millisecond. If that happens you'll not be able to discriminate between those two fragments.

The two general approaches I've seen for handling this particular issue in tests is to either manually specify the timestamps on write, or insert a short 1-2ms sleep.

davisp (approved):
+1 overall though we'll want to consider that timestamp issue. It may not be biting us now if the proptests are slow enough, but its one of those lurking bugs to at least have a guard for (i.e., add a check on the number of opened fragments and panic if its more than 1).

tiledb/api/src/array/mod.rs
```diff
@@ -224,27 +226,22 @@ impl Array {
...
 
-        Ok(Array {
-            context: context.clone(),
-            uri: uri.as_ref().to_owned(),
-            raw: RawArray::Owned(array_raw),
-        })
+    /// Prepares an array to be "re-opened". Re-opening the array will bring in any changes
+    /// which occured since it was initially opened. This also allows changing configurations
+    /// of an open array, such as the timestamp range.
+    pub fn refresh(self) -> ArrayOpener {
```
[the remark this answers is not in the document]
rroelke: Eh, I'll switch to `reopen` at least since it's a searchable term in the C docs.

tiledb/api/src/query/write/strategy.rs
rroelke: Hm, there's a few assumptions built in then:
1) the fragment list is sorted on timestamp, so that `fi.get_fragment(num_fragments - 1)` always gives a handle to the one we just wrote
2) things are slow enough that the previous fragment's `timestamp_min` is less than the most recent's `timestamp_max`

Inserting a sleep seems safe though I'd be quite surprised if we hit #2 since each loop of the test is quite hefty

PR #110, opened 2024-05-31
Use 'sm.var_offsets.extra_element=true' for write queries

rroelke:
`tiledb-rs` queries use extra offsets for read mode, but not write mode.  This creates some confusion when interacting with raw buffers - a buffer cannot be passed from read query to write query, it has to be adjusted.  It also creates some difficulty with testing and conversion to arrow.

This pull request synchronizes read and write queries in their usage of the extra offset, which simplifies a lot of the testing code and will further pay off in upstream crates and repositories.


davisp (approved):
+1

davisp:
Err, though we should fix that segfault before merging...

rroelke:
Yep, this depends on https://github.com/TileDB-Inc/TileDB/pull/5033 in core, and I expect that this will pass once that is merged (based on my experience on my laptop)

PR #111, opened 2024-06-04
Alternative strategy for a vector of records

rroelke:
The strategy produced by `proptest::collection::vec` shrinks very slowly, lopping records off from the end one at a time.  It also stores a value tree for each element of the `Vec`, and shrinks those when the minimum set of elements has been determined.

This strategy is easy to use but is not really right for some of our use cases.  For example, if we want to write a large number of records into tiledb and the test fails, then shrinking the input data one record at a time is hardly sensible.  And shrinking the individual elements, while not worthless, is also not worth the state overhead.

This pull request takes the strategy already written to reason about sparse write inputs, and generalizes it to a new `Records` trait.

A simple experiment shows that on large inputs, this new strategy shrinks down to nothing significantly faster (~2 minutes vs. ~14 minutes in one experiment in a debug build, which while unreliable for a performance comparison is nonetheless the manner in which we run most of our testing).

### DIFF
--- a/tiledb/test-utils/src/strategy/records.rs
+++ b/tiledb/test-utils/src/strategy/records.rs
@@ -10,6 +10,10 @@ use proptest::prelude::*;
 use proptest::strategy::{NewTree, ValueTree};
 use proptest::test_runner::TestRunner;
 
+/// Create a strategy to generate `Vec`s containing elements drawn from `element` and with a size
+/// range given by `size`.  In contrast to `proptest::collection::vec`, value trees produced by
+/// this strategy do not attempt to shrink any of the elements of the vector, instead shrinking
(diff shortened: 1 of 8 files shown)


tiledb/test-utils/src/strategy/records.rs
```diff
@@ -0,0 +1,462 @@
...
+                        $T.filter(subset),
+                    )+
+                )
+            }
+        }
+    };
+}
+
+// this is what std does for identifiers, who knows why
+tuple_impls_record!(E, D, C, B, A, Z, Y, X, W, V, U, T);
```
davisp, line 129: Huh.

davisp (approved):
+1

PR #112, opened 2024-06-05
Various metadata API changes and improvements in support of SC-48994

rroelke:
SC-48994 is about adding a DataFusion TableProvider which scans, and later maybe inserts, metadata.

The tables PR which depends on this is [here](https://github.com/TileDB-Inc/TileDB-Tables/pull/17).

### DIFF
--- a/tiledb/api/src/metadata.rs
+++ b/tiledb/api/src/metadata.rs
@@ -253,6 +253,7 @@ pub mod strategy {
     use proptest::collection::{vec, SizeRange};
     use proptest::prelude::*;
 
+    use crate::datatype::strategy::DatatypeContext;
     use crate::datatype::LogicalType;
 
     pub struct Requirements {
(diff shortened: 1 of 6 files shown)


tiledb/api/src/metadata.rs
```diff
@@ -20,9 +20,6 @@ pub enum Value {
     Int64Value(Vec<i64>),
     Float32Value(Vec<f32>),
     Float64Value(Vec<f64>),
-    //StringAsciiValue(CString),
```
davisp, line 23: Is there no support for strings or is the idea to return then as a Vec<u8> and rely on external bits to do the conversion?

davisp (approved):
+1. Just had a minor thought. If we've just not added String support yet, we'll want to make sure and get a ticket for that. My hunch is that a major use of metadata is for storing JSON blobs and asking around seems to at least confirm that Strings are *mostly* but not always the datatype used.

tiledb/api/src/metadata.rs
rroelke: That seems to be how we've generally done things, string goes in as bytes and the user is responsible for checking UTF-8 when it comes out.

Since core doesn't validate UTF-8 I'm not sure we really have another option.  We could add a `From<String>` or something though so we can move the contents of a string into metadata

PR #113, opened 2024-06-07
Aggregate API support

abigalekim:
This pull request adds support for running global aggregates using the query API.

We add a new trait `AggregateQueryBuilder` which is implemented by queries that are compatible with aggregates.  Currently, this is only the base query builder.  This means that either all of the query results are aggregate functions, or none of them is; and this assumption is built into the implementation.

`AggregateQueryBuilder` provides functions for adding global aggregate results.  The result of a global aggregate is a single value, so instead of providing buffer space to write the value to, the query _result_ is the returned value.  The user must request a type of their result and compatibility with the anticipated result type is checked.  This is demonstrated in the `aggregates` example.

In addition to that, we add a new crate `queries` which is intended to be a starting point for adding new (or moving existing, eventually) query adapters.  Here we add a trait `AggregateQueryBuilderExt` which provides methods for returning a `PhysicalValue` to hold the result instead of a different value.  We demonstrate this API in the `aggregate-adapters` example.

Note that we expect to use the `AggregateExt` version in tables.

# Testing

Intermediate states of this pull request contain property-based testing that we can run aggregate functions of all kinds on all field types.  This exposed a number of bugs in core which slogged the development here due to working around them.  As such I have removed these tests and added some specific examples instead to `query/read/aggregate/mod.rs`.  It does look like there are a few artifacts of this in some of the `strategy` modules; I have left them in for now but am happy to remove them upon request.


tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +375,104 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+    }
+}
+
+pub enum AggregateType {
+    Count,
+    Sum
+    // agg_type : AggregateType , attr_name : Option<String>
+}
+
+pub trait AggregateBuilder : QueryBuilder {
```
rroelke, line 445: `/// Trait for query types which can have an aggregate channel placed on top of them`

tiledb/api/examples/aggregates.rs
```diff
@@ -0,0 +1,165 @@
...
+        .layout(tiledb::query::QueryLayout::RowMajor)?
+        .apply_aggregate::<u64>(AggregateType::Count, None)?
+        .start_subarray()?
+        .add_range("rows", &[1i32, 2])?
+        .add_range("columns", &[2i32, 4])?
+        .finish_subarray()?
+        .build();
+
+    let results = query.execute()?;
+    println!("{}", results);
```
rroelke, line 126: I think it would be slightly more illustrative to return a strongly-typed result instead and have the caller print it.
(That way the example code shows that this just straight-up returns a u64)

tiledb/api/examples/aggregates.rs
```diff
@@ -0,0 +1,165 @@
...
+}
+
+fn main() {
+    if !array_exists() {
+        create_array().expect("Failed to create array");
+    }
+    write_array().expect("Failed to write array");
+    //get_count().expect("Failed to count array");
+    get_sum().expect("Failed to sum array");
+}
```
rroelke, line 165: When the aggregate builder supports doing multiple aggregate calls (which the C API does support, right?) then please add another function to this example which demonstrates doing so

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+    base: B,
+    agg_str: CString,
+    attr_str: Option<CString>,
+    attr_type: PhantomData<T>,
+}
+
+pub struct AggregateReader<Q, T> {
+    base: Q,
+    agg_str: CString,
+    _attr_str: Option<CString>,
```
rroelke, line 397: Make sure to leave a comment as to why this is unused

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+            ReadStepOutput::Final(_) => self.data,
+            ReadStepOutput::Intermediate(_) => unreachable!(),
+            ReadStepOutput::NotEnoughSpace => unreachable!(),
+        };
+
+        Ok(ReadStepOutput::Final(return_val))
+    }
+}
+
+#[derive(PartialEq)]
```
rroelke, line 479: I reckon you can have Eq in here too.  ANd I'd suggest Debug and Clone.
And for proptest you can also derive `Arbitrary`, that's a little trickier but there are other examples for things like `TileOrder` probably

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+pub enum AggregateType {
+    Count,
+    Max,
+    Mean,
+    Min,
+    NullCount,
```
rroelke, line 485: NullCount as in, count the number of null values?

tiledb/api/examples/aggregates.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    let array = tiledb::Array::open(
+        &tdb,
+        QUICKSTART_DENSE_ARRAY_URI,
+        tiledb::array::Mode::Read,
+    )?;
+
+    let mut query = tiledb::query::ReadBuilder::new(array)?
+        .layout(tiledb::query::QueryLayout::RowMajor)?
+        .apply_aggregate::<u64>(AggregateType::Count, None)?
```
rroelke, line 118: What do you think of making some trait-default methods like `fn count(self)` for which this is the implementation?
I think that will assist the flow of the user code.
It will also offer you a good place to put doc comments with examples for the individual functions.

(this version `apply_aggregate` is still useful of course, we can definitely deploy it in proptest and tables)

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+            ReadStepOutput::NotEnoughSpace => unreachable!(),
+        };
+
+        Ok(ReadStepOutput::Final(return_val))
+    }
+}
+
+#[derive(PartialEq)]
+pub enum AggregateType {
+    Count,
```
rroelke, line 481: If the variants are to be in the public API I'd like to see doc comments for each explaining what the input/output is and when the function is allowed to be used.  This is less important if you do the `fn count`, `fn max`, etc approach, which is my preference.

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+    Min,
+    NullCount,
+    Sum,
+}
+
+pub trait AggregateBuilderTrait: QueryBuilder {
+    fn apply_aggregate<T>(
+        self,
+        agg_type: AggregateType,
+        name: Option<String>,
```
rroelke, line 493: It's a little awkward since Count is the only one that doesn't require a name.  I think it would be better to put the field name inside of the enum, e.g.
```
enum AggregateType {
    Count,
    Max(String),
    ...
}
```
instead of a type and separate option.

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+    agg_str: CString,
+    _attr_str: Option<CString>,
+    data : T,
+    data_size : u64
+}
+
+impl<B, T> QueryBuilder for AggregateBuilder<B, T>
+where
+    B: QueryBuilder,
+    T : Default,
```
rroelke, line 405: When we want to say that a type is supposed to be something that comes out of a tiledb query, I think we're best off using `T: PhysicalType` from our `datatype` module.

Is that more restrictive than is really needed here?  Yes.  But I think it's a better expression of intent.

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+
+        context.capi_call(|ctx| unsafe {
+            ffi::tiledb_query_set_data_buffer(
+                ctx, cquery, agg_c_ptr, c_bufptr, c_sizeptr,
+            )
+        })?;
+
+        let base_result = self.base.step()?;
+
+        // Run the query in a loop until you get the final result
```
rroelke, line 468: Comment is misleading, there's no loop. Better to explain why the intermediate/not enough space results are unreachable.

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+        let return_val = match base_result {
+            ReadStepOutput::Final(_) => self.data,
+            ReadStepOutput::Intermediate(_) => unreachable!(),
```
rroelke, line 471: `unreachable!` has an optional argument for context which is a format string. Please fill it in just in case something does happen so we have a bit of context from a bug report.

tiledb/api/src/query/read/mod.rs
```diff
@@ -374,6 +384,262 @@ pub trait ReadQueryBuilder<'data>: QueryBuilder {
...
+                ctx,
+                default_channel,
+                c_agg_name,
+                agg_operation,
+            )
+        })?;
+
+        Ok(AggregateBuilder::<Self, T> {
+            base: self,
+            agg_str: agg_name,
```
rroelke, line 636: Is the agg name used after the channel initialization? Is it needed to keep it around in the same way that the attribute name is?

rroelke (commented):
I know this is still in draft state but I felt inspired to take a look and give some early feedback. I hope it is helpful and not annoying that I have done so before you were fully ready

tiledb/api/examples/aggregates.rs
abigalekim: it doesn't yet :( but I definitely will when I do support that!

tiledb/api/src/query/read/mod.rs
abigalekim: yes!

tiledb/api/examples/aggregates.rs
abigalekim: I think that would be a good idea!

tiledb/api/src/query/read/mod.rs
abigalekim: yes. Core TileDB stores the aggregate name in an unordered map:
/** Aggregates for the default channel, by output field name. */
  std::unordered_map<std::string, shared_ptr<IAggregator>>
      default_channel_aggregates_;

tiledb/api/src/query/read/mod.rs
abigalekim: fixed!

tiledb/api/src/query/read/mod.rs
abigalekim: done!

tiledb/api/src/query/read/aggregate.rs
```diff
@@ -0,0 +1,665 @@
...
+/// Trait for query types which can have an aggregate channel placed on top of them.
+pub trait AggregateBuilderTrait: QueryBuilder {
+    fn apply_typed_aggregate<T: 'static>(
+        self,
+        agg_type: AggregateType,
+    ) -> TileDBResult<AggregateTypedBuilder<Self, T>> {
+        let expected_type =
+            aggregate_type(&agg_type, &self.base().array().schema()?)?;
+        if !expected_type.is_compatible_type::<T>() {
+            return Err(TileDBError::InvalidArgument(anyhow!(
```
rroelke, line 436: This should be `Error::TypeMismatch` instead of InvalidArgument

tiledb/api/src/query/read/aggregate.rs
```diff
@@ -0,0 +1,665 @@
...
+            Datatype::Int64 => AggregateEnumBuilder::Int64(
+                self.apply_typed_aggregate::<i64>(agg_type)?,
+            ),
+            Datatype::Float32 => AggregateEnumBuilder::Float32(
+                self.apply_typed_aggregate::<f32>(agg_type)?,
+            ),
+            Datatype::Float64 => AggregateEnumBuilder::Float64(
+                self.apply_typed_aggregate::<f64>(agg_type)?,
+            ),
+            _ => unreachable!(
```
rroelke, line 643: This looks reachable to me, what if you do this on a `Datatype::StringAscii` attribute?  Or a timestamp.  This needs to work generically on the physical type, which fortunately `physical_type_go!` exists to help you do.

I do see a possible concern though - what do min/max do in core for a var-sized attribute?  Are those valid? Do we need some other kind of result holder that also can hold a `Vec` of its datatype?

tiledb/api/src/query/read/aggregate.rs
```diff
@@ -0,0 +1,665 @@
...
+
+    fn max<T: 'static>(
+        self,
+        attr_name: String,
+    ) -> TileDBResult<AggregateTypedBuilder<Self, T>> {
+        self.apply_typed_aggregate::<T>(AggregateType::Max(attr_name))
+    }
+}
+
+pub trait AggregateEnumBuilderTrait: QueryBuilder {
```
rroelke, line 602: the bounds will be simpler if you just require `AggregateBuilderTrait` for the whole trait.

tiledb/api/src/query/read/aggregate.rs
```diff
@@ -0,0 +1,665 @@
...
+                self.apply_typed_aggregate::<f64>(agg_type)?,
+            ),
+            _ => unreachable!(
+                "aggregate_type function should return a numeric type."
+            ),
+        };
+        Ok(agg_builder)
+    }
+}
+
```
rroelke, line 650: I definitely want to see a `mod tests` in here somewhere which covers the error conditions, as there are quite a few of them.

rroelke (changes requested):
Made another pass, the design looks good so the comments are largely in the details.

In addition to the in-line comments I think the `EnumThing` naming convention should be changed.  I was looking at other examples and we aren't super super consistent, but the most prominent convention is actually that we have `struct MyThing<T>` with the generic and then `enum TypedMyThing`.  It would be better to follow that convention.

tiledb/api/examples/aggregates.rs
```diff
@@ -0,0 +1,297 @@
...
+        _ => unreachable!("Wrong return type!"),
+    };
+
+    println!("{}", min_res);
+    println!("{}", max_res);
+
+    Ok(())
+}
+
+fn get_min_max_half() -> TileDBResult<()> {
```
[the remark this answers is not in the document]
abigalekim: oh this was just for testing to see if i could return an enum and a typed result in the same aggregate query. it should be moved to a test and has been deleted

tiledb/api/src/query/read/aggregate.rs
abigalekim: unfortunately this approach doesn't work bc the match arms will have different types. Also I think aggregates just support numerics for now.

tiledb/api/src/query/read/aggregate.rs
```diff
@@ -0,0 +1,665 @@
...
+            };
+            Ok(enum_result)
+        })
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum AggregateType {
+    Count,
+    Max(String),
```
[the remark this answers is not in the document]
abigalekim: can we use it in datafusion if it is private?

tiledb/api/src/query/read/aggregate.rs
abigalekim: This takes a string so I just used typeID. Is there a better way of converting a type into a string?

tiledb/api/src/query/read/aggregate.rs
```diff
@@ -0,0 +1,665 @@
...
+        })?;
+
+        let base_result = self.base.step()?;
+
+        // There are no intermediate results for aggregates since the buffer size should be one
+        // element (and therefore, no space constraints).
+        let (return_val, base_q) = match base_result {
+            ReadStepOutput::Final(base_q) => (self.data, base_q),
+            ReadStepOutput::Intermediate(_) => {
+                unreachable!("Aggregate step function.")
```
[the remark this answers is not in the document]
abigalekim: This doesn't work since the base result type is not required to implement debug or display. How do we mitigate this problem?

tiledb/api/src/query/read/aggregate.rs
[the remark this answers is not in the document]
rroelke: I think no matter what, we're going to have some kind of match on a datafusion expression turning into some kind of way of doing aggregation... but you're right that turning that into this enum is probably the easier route.  I've got a bit too much of the example code on my brain.

I'd probably just change this to `AggregateOp` then, as `Type` is something of an overloaded term here.

tiledb/api/src/query/read/aggregate.rs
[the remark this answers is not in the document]
rroelke: `ReadStepOutput` derives Debug, so it should be Debug when its type parameters are.

As far as this `unerachable!` goes, that's a fair point. But I think it's ok to put the needed bounds in there.
- the bound on `T` should be PhysicalType rather than Copy, and thus implements Debug. IIRC I left this comment previously about a different bound but it applies here too.
- `<Q as ReadQuery>::Intermediate` and `<Q as ReadQuery>::Final` are always going to be Debug in practice anyway since aggregates can only be put on top of each other and the base query.

tiledb/api/src/query/read/aggregate.rs
rroelke: Why would the match arms all have different types?  You could do `AggregateEnumBuilder::from` to get around that


Even if not this still looks reachable.  If you try to do min or max of a datetime type then where is the point of error?

rroelke: Resolution: `phyiscal_type_go!`

tiledb/api/src/query/read/aggregate.rs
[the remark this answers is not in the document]
rroelke: Resolution: changed to `AggregateFunction`

rroelke:
@davisp would you like to take a look at this now?  I have updated the testing to not use the proptests that were tripping on core bugs; and I think everything else is in order.

tiledb/queries/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "tiledb-queries"
```
davisp, line 2: I'm not sure I understand why this is a separate crate. Also, the name doesn't strike me as super awesome given how general it is compared to what we've got here.

tiledb/api/src/query/read/aggregate/mod.rs
```diff
@@ -0,0 +1,1054 @@
...
+                        | Datatype::Int32
+                        | Datatype::Int64 => Some(Datatype::Int64),
+                        Datatype::UInt8
+                        | Datatype::UInt16
+                        | Datatype::UInt32
+                        | Datatype::UInt64 => Some(Datatype::UInt64),
+                        Datatype::Float32 | Datatype::Float64 => {
+                            Some(Datatype::Float64)
+                        }
+                        _ => None,
```
davisp, line 105: Is this dictated by core or am I missing a Rust reason for the coercion to these types?

tiledb/api/src/query/read/aggregate/mod.rs
```diff
@@ -0,0 +1,1054 @@
...
+        })?;
+
+        Ok(AggregateBuilder::<T, Self> {
+            base: self,
+            handle,
+            field_type: PhantomData,
+        })
+    }
+
+    /// Adds a request for the number of results which satisfy the query predicates.
```
davisp, line 425: Does core return a count of matching results when given a query condition? I only vaguely paid attention when this was being implemented and I don't remember if that was supported or if it was just about "number of cells in the subarray ranges provided".

tiledb/api/src/datatype/physical.rs
```diff
@@ -330,3 +331,125 @@ where
...
     }
 }
+
+/// Represents a dynamically typed single physical value.
+///
+/// [PhysicalValue] holds the bits which correspond to a single value of a logical data type.
+/// For a given logical data type, [PhysicalValue] will always be constructed with the
+/// primitive numerical type of the same bit width, signed-ness, and precision.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum PhysicalValue {
```
davisp, line 341: Does core provide any aggregates for string fields? I know we store some stuff in fragment metadata for string fields, but obvious sum aggregates wouldn't work there.

I don't want to derail this PR by going back to add string support if so, but I just wanted to note whether this is yet another case where we have an API where we basically need a "any physical type or a string" sort of enumeration. We seem to be collecting variations on this theme and at some point we'll probably want to go back and unify them all.

tiledb/api/src/query/read/aggregate/mod.rs
```diff
@@ -0,0 +1,1054 @@
...
+        )?;
+
+        let mut q = ReadBuilder::new(a.for_read()?)?
+            .min::<i32>("a")?
+            .max::<i32>("a")?
+            .build();
+        let (a_max, (a_min, _)) = q.execute()?;
+
+        // This is deliberately a wrong result.
+        // We capture it here to track SC-54468.
```
davisp, line 1019: Super minor nit, I really like these "add tests that will fail when bug is fixed" approach. I think we should consider standardizing the `sc_#####_` prefix for tests that exist for this reason.

davisp (approved):
I had a couple notes, but nothing major to really add or change.

I think we should file two follow up stories though:

First, lets figure out the current level of string support in min/max type aggregates. The  C API does not appear to support things, but I didn't dig too far into them to see if I was just missing how that was supposed to work. As noted, we've also got this whole pattern of "any physical type or String" situation in a number of places. We'll probably want to also unify those at some point.

Secondly, for "return query results with aggregates" possibility, it seems like core would support that but our current implementation doesn't. That's fine for me since that behavior seems a bit niche, especially given the use cases we expect for these. But it might make for a decent "first ticket" sort of thing to rework the current approach into something that does support the possibility.

Otherwise, everything looks great to me. The only major thing was that tiledb-queries crate which I'm -0 on. Its not my favorite but I'm not going to veto it since we can always refactor it later.

tiledb/api/src/query/read/aggregate/mod.rs
rroelke: I think this is dictated by core. But not in a way that's exposed by the API.

If it is then we should find the code that dictates it and leave a note here.

tiledb/api/src/query/read/aggregate/mod.rs
rroelke: Will do

tiledb/api/src/datatype/physical.rs
rroelke: From discussion:

yes, it does, one must set an offset buffer as they usually do.

This will go in a follow-up story.

tiledb/queries/Cargo.toml
rroelke: I was envisioning that we might eventually migrate the other query adapters into this crate and leave the API crate with just the barest form of "please fill in my buffer" query API.  Hence the choice of generic name and crate.

rroelke: Maybe I should re-name it to `tiledb-query-adapters`?

tiledb/api/src/datatype/physical.rs
```diff
@@ -330,3 +331,125 @@ where
...
+                        _,
+                        Err(Error::physical_type_mismatch::<$ty, DT>())
+                    )
+                }
+            }
+        }
+    };
+}
+
+physical_value_traits!(i8, Int8);
```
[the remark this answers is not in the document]
rroelke: Sure

tiledb/api/src/query/read/aggregate/mod.rs
rroelke: From what I can tell:

The C API reasons about buffer sizes, not about the type of the pointee.  So the output buffer validation code basically does the same thing.  Each aggregate has its own way of calling a function in `output_buffer_validator.cc`.

Then each aggregate has its own way of copying its intermediate result to the final output.  Sum has `sum_type_data`, Mean just always writes double.

All this to say, I think there's a few references we can add here, but not a single function

tiledb/queries/Cargo.toml
davisp: I'd be +1 for that rename.

tiledb/api/src/datatype/physical.rs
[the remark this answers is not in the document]
rroelke: 5b1cf1574

tiledb/queries/Cargo.toml
rroelke: 34ad527

tiledb/api/src/datatype/physical.rs
rroelke: https://app.shortcut.com/tiledb-inc/story/54898/rs-add-support-for-running-min-max-aggregates-on-string-fields

tiledb/api/src/query/read/aggregate/mod.rs
rroelke: 4c1e4c0165

tiledb/api/src/query/read/aggregate/mod.rs
rroelke: 22126ded

tiledb/api/src/query/read/aggregate/mod.rs
rroelke: Confirmed via example that changing the subarray changes aggregate results

rroelke (approved):

PR #117, opened 2024-06-11
Remove Option return from Context::object_type

rroelke:
This small pull request is a very minor usability enhancement.  The function `Context::object_type` returns an `ObjectType` for whatever it finds at a given URI.

This method currently returns `Option`... but not for any documented reason.  And the one you would expect, that an object is not found at all, is actually an error from inside core.

Why do we return `Option`?  Because of a possible conversion failure from the C value to the `ObjectType` enum.  But that should be an internal error, as indicated the method this PR switches to using instead.


abigalekim (approved):
Looks good to me!

PR #118, opened 2024-06-11
Reduce DEFAULT_MAX_SPARSE_TILE_CAPACITY to reduce max size of proptest sparse tiles

rroelke:
I believe this is what has been responsible for the spurious failures in https://github.com/TileDB-Inc/TileDB-Tables/pull/16 .  When the schema strategy chooses a very large sparse tile capacity, we allocate huge amounts of memory in the query writer, which can lead to OOM in any environment.  The previous limit apparently led to OOMs frequently enough in CI to prevent merging.

This pull request syncs the default limit on sparse tile capacity with the default upper bound on dense array domains.  This bound is much less and should not result in OOMs.

### DIFF
--- a/tiledb/api/src/array/schema/mod.rs
+++ b/tiledb/api/src/array/schema/mod.rs
@@ -410,6 +410,10 @@ impl Schema {
         ArrayType::try_from(c_atype)
     }
 
+    /// Returns the sparse tile capacity for this schema,
+    /// i.e. the number of cells which are contained in each tile of a sparse schema.
+    /// If this is a dense array schema, the value returned is
+    /// not used by tiledb.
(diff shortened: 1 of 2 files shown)


rroelke:
Do we consider this to be a bug in core?  That space for the whole tile is allocated, no matter how large the tile is?

Seems sort of low priority to fix, I don't imagine tiles that large are really right for any applications we know of at this time

davisp (approved):
+1

I doubt core would say this is a bug vs expected behavior. I'm pretty sure we always load an entire tile at a time because of how filters work and doing it "incrementally" would just be the moral equivalent of "make your capacity smaller".

PR #120, opened 2024-06-14
Update to arrow 52.0

rroelke:
Datafusion version 39 uses arrow version 52, so we must update `tiledb-rs` to match first.


davisp (approved):
+1

PR #121, opened 2024-06-14
chore: [sc-49541] [rs] add memory limit to default allocation strategy for query read buffer

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/49541

This should fix, or at least significantly diminish the probability of, the "could not allocate memory" issues we observed in `tables` CI.

### DIFF
--- a/tiledb/api/src/array/attribute/mod.rs
+++ b/tiledb/api/src/array/attribute/mod.rs
@@ -728,6 +728,34 @@ mod test {
     use crate::filter::list::Builder as FilterListBuilder;
     use crate::filter::*;
 
+    /// Test what the default values filled in for `None` with attribute data are.
+    /// Mostly because if we write code which does need the default, we're expecting
+    /// to match core and need to be notified if something changes or we did something
+    /// wrong.
+    #[test]
+    fn attribute_defaults() {
+        let ctx = Context::new().expect("Error creating context instance.");
+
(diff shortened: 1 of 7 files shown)


tiledb/api/src/array/attribute/mod.rs
```diff
@@ -728,6 +728,34 @@ mod test {
...
+            assert_eq!(CellValNum::single(), attr.cell_val_num().unwrap());
+        }
+        {
+            let spec = AttributeData {
+                name: "xkcd".to_owned(),
+                datatype: Datatype::StringAscii,
+                ..Default::default()
+            };
+            let attr = spec.create(&ctx).unwrap();
+            assert_eq!(CellValNum::single(), attr.cell_val_num().unwrap());
```
davisp, line 755: Huh. I trust CI on this, but I'm surprised by StringAscii defaulting to cell_val_num 1. Might be something worth filing a ticket for core on.

davisp (approved):
+1

tiledb/api/src/array/attribute/mod.rs
rroelke: I was also surprised, I wrote the test with `CellValNum::Var` expecting it to pass and it did not!

You do consider that a bug? I can see it from both sides

PR #122, opened 2024-06-14
Add test utility to run a callback as a signal handler

rroelke:
This was useful in debugging the recent tables CI issues.  When Rust fails to allocate memory, it terminates the process with SIGABRT which does not bubble up to proptest to print the current test case.  Using this utility I was able to register a callback to print the current test case in response to that SIGABRT, leading to a reproducer I could inspect on my laptop.

It's certainly not production ready, as some of the comments indicate, but I'd like to merge it in case it proves useful again in the future.

### DIFF
--- a/tiledb/test-utils/src/lib.rs
+++ b/tiledb/test-utils/src/lib.rs
@@ -1,6 +1,9 @@
 extern crate proptest;
 
+pub mod signal;
 pub mod strategy;
 pub mod uri_generators;
 
+pub use nix::sys::signal::{SigHandler, Signal};
+pub use signal::*;
 pub use uri_generators::{get_uri_generator, TestArrayUri};
--- a/tiledb/test-utils/src/lib.rs
+++ b/tiledb/test-utils/src/lib.rs
@@ -4,6 +4,5 @@ pub mod signal;
 pub mod strategy;
 pub mod uri_generators;
 
-pub use nix::sys::signal::{SigHandler, Signal};
 pub use signal::*;
 pub use uri_generators::{get_uri_generator, TestArrayUri};
--- a/tiledb/test-utils/src/lib.rs
+++ b/tiledb/test-utils/src/lib.rs
@@ -1,8 +1,11 @@
 extern crate proptest;
 
+#[cfg(feature = "signal")]
 pub mod signal;
 pub mod strategy;
 pub mod uri_generators;
 
+#[cfg(feature = "signal")]
 pub use signal::*;
+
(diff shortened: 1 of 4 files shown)


rroelke:
Assertion failures inside core manifest as SIGABRT so this actually might event be something we want to put in our tests by default

tiledb/test-utils/src/signal.rs
```diff
@@ -0,0 +1,232 @@
...
+        RawSignalCallback(unsafe { std::mem::transmute::<_, _>(ptr) })
+    }
+}
+
+unsafe impl Send for RawSignalCallback {}
+
+const SIGNAL_CALLBACK_DEFAULT: Mutex<Option<RawSignalCallback>> =
+    Mutex::new(None);
+static SIGNAL_CALLBACKS: [Mutex<Option<RawSignalCallback>>; 32] =
+    [SIGNAL_CALLBACK_DEFAULT; 32];
```
davisp, line 48: Is this documented as a constant limit across all of the various *nix platforms? It appears to be true on Darwin, but I'm not sure if there's a BSD out there with more than 32 signals.

tiledb/test-utils/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 extern crate proptest;
 
+pub mod signal;
```
davisp, line 3: We should probably make this a `#[cfg(!windows)]` or whatever. I know we're not building there yet, but it'll be a requirement soonishly. Or perhaps a feature flag as it seems like the sort of thing we'd only enable as a tool for debugging something that's failing? Not sure on that one. I might be nice to have something that attempts to print a stack on any signal encountered during test runs.

davisp (approved):
I'm on board having this around, but we should either gate it on a config setting or behind a feature. I've not got a strong preference which route to go with that.

tiledb/test-utils/src/signal.rs
rroelke: Probably not, but it is a limit inside the `nix` crate.
https://docs.rs/nix/latest/nix/sys/signal/enum.Signal.html

rroelke:
CI passed which means clippy is not running on the `test-utils` crate, I ran it locally and there were some things turned up (now fixed).

tiledb/test-utils/src/lib.rs
rroelke: Resolution: "signal" feature, and the interior implementation is behind `#[cfg(not(windows))]` so that the API is the same regardless of platform.

PR #125, opened 2024-06-21
Split fn_typed into logical_type_go and physical_type_go and add example

rroelke:
`fn_typed` was our earliest foray into this sort of meta-programming and it predates what has become our preferred style.  Additionally, almost all usages of it immediately bound a new type identifier `type DT = <LT as LogicalType>::PhysicalType`.

This pull request addresses both of those by splitting `fn_typed` into `logical_type_go` and `physical_type_go`, whose purposes are hopefully clear from the names.


davisp (approved):
+1

PR #127, opened 2024-06-29
lint map => inspect

rroelke:


davisp (approved):
+1 after a rebase with green CI.

PR #131, opened 2024-07-02
Consistent output directory for examples

rroelke:
As written each of our examples creates its array in the current working directory from which you run the example. This clutters the workspace.  This pull request updates the examples to instead write to a consistent output location where we can place a `.gitignore`.

Along the way we fix `Context::object_type` (again - see https://github.com/TileDB-Inc/tiledb-rs/pull/117) to return `None` for paths which do not exist in the absence of I/O errors.


davisp (approved):
+1

PR #134, opened 2024-07-09
feat: [sc-50382] [tables] or [rs] implement accumulating a subarray from the intersection of predicate ranges

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/50382

This is a follow-up to #133 which implements the functionality we're *most* interested in, which is the intersection of two subarrays rather than just the intersection of a subarray and a range.

If we have a predicate `(a OR b) AND (c OR d)`, then range analysis will (eventually) give us something like `((dim IN [100, 200]) OR (dim IN [300, 400])) AND ((dim IN 150, 150) OR (dim in [350, 450]))`.  This reduces to `(dim IN [150, 200]) OR (dim IN [300, 350])`.  This pull request implements that reduction and verifies it with property-based testing.

Along the way we also fix the previous function `intersect_ranges`'s handling of the "empty range means everything" special case.

### DIFF
--- a/tiledb/api/src/query/subarray.rs
+++ b/tiledb/api/src/query/subarray.rs
@@ -2,6 +2,7 @@ use std::marker::PhantomData;
 use std::ops::Deref;
 
 use anyhow::anyhow;
+use itertools::Itertools;
 
 use crate::array::Schema;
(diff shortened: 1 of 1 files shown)


davisp (approved):
+1

PR #136, opened 2024-07-10
feat: [sc-45521] CI does not run tests if linter fails

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/45521

The title is not quite right, it should say "tests do not run if linter fails".

This pull request fixes the issue in the title by splitting the linting off into a separate workflow.


davisp (approved):
+1

PR #138, opened 2024-07-17
SC-51253

svilen-mihaylov-db:
[SC-51253](https://app.shortcut.com/tiledb-inc/story/51253/tweak-the-makefile-which-produces-lists-of-implemented-covered-and-desired-to-implement-todo-functions)

Tweak the makefile which produces lists of implemented (covered) and desired to implement (TODO) functions

### DIFF
--- a/tiledb/sys/Makefile
+++ b/tiledb/sys/Makefile
@@ -10,11 +10,20 @@ generated.rs:
     -- -I/opt/tiledb/include
 
 generated-functions.txt: generated.rs
+    # Which functions are exponsed by the C++ API.
 	rg -o --no-line-number --no-filename '\bfn\s+tiledb_[^(]+' generated.rs | sort > generated-functions.txt
 
 covered-functions.txt:
+    # Which functions are already implemented.
 	rg -o --no-line-number --no-filename '\bfn\s+tiledb_[^(]+' src/*.rs | sort > covered-functions.txt
 
-todo: generated-functions.txt covered-functions.txt
-	comm -23 generated-functions.txt covered-functions.txt > todo.txt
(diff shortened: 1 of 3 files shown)


tiledb/sys/Makefile
```diff
@@ -10,11 +10,20 @@ generated.rs:
...
-todo: generated-functions.txt covered-functions.txt
-	comm -23 generated-functions.txt covered-functions.txt > todo.txt
+ignored-functions.txt:
+    # Which functions we do not wish to implement.
+	rg -o --no-line-number --no-filename '\bfn\s+tiledb_[^(]+' ignored.rs | sort > ignored-functions.txt
+
+todo: generated-functions.txt covered-functions.txt ignored-functions.txt
+	comm -23 generated-functions.txt covered-functions.txt > not-covered.txt
+	comm -23 not-covered.txt ignored-functions.txt > todo.txt
+	comm -12 covered-functions.txt ignored-functions.txt > ignored-but-covered.txt
```
svilen-mihaylov-db, line 27: Note he have some functions which are ignored but are also covered:
fn tiledb_array_get_non_empty_domain
fn tiledb_filter_option_from_str
fn tiledb_filter_option_to_str

davisp: Might be a good idea to add logic checking for these in the api-coverage tool then.

davisp (approved):
I'm fine with this change, though it might make more sense to abandon this Makefile stuff and just update the api-coverage tool with a terminal output format option instead of HTML tables. I think I've seen rust things for TTY detection to even make it automagical. You definitely don't have to do that instead of this though. Just a thought for the future.

PR #139, opened 2024-07-22
WIP: Create a static build of tiledb-rs

davisp:
The eventual goal here is to allow Rust bindings to be built statically so that there's no dependency on a libtiledb.so/dll. Although we'll need both approaches for things like Python bindings.


rroelke:
What kind of CI updates do we want here?  Should we run all the tests on both the static and dynamic builds?  Certainly we ought to build both them both at the least.

tiledb/sys/build.rs
```diff
@@ -1,11 +1,172 @@
+#[cfg(feature = "static")]
+mod tiledb {
+    use std::collections::HashSet;
+
+    pub fn configure() {
```
rroelke, line 5: I understand part of what you are doing here, some comments on the more obscure parts will help

`update_repository` is very clear, find the tiledb repository in `$OUT_DIR/repos` (I assume `$OUT_DIR` ends up as `target`), if it exists then pull to latest of whatever the current branch is.  If it does not exist then clone it.  Is this always the `dev` branch?

`copy_link_source`, I have little idea what is going on here. I'm not acquainted with cmake so I don't really see how `link_info.cc` factors in.  If we compile this with a static archive then wouldn't the unused symbols from the archive be omitted from the final compilation output?

`build_tiledb` seems clear enough

`write_linker_commands`, the assumption is that building tiledb produced a file named `link_info.txt` which is a list of the libraries needed to link with, and this script prints a command line argument for cargo to statically link with those libraries

davisp: For now, just consider this a temporary hack that may or may not outlive this draft PR state. I may or may not add a static CI depending on what we find to be the easiest way to get a CLI into folks' hands. I'm currently poking at an alternative solution that continues to use the libtiledb.so/dll/dylib using an RPATH set to the current directory.

You've mostly got where I was going on the static build stuff. The CMake bits you're missing is that I set the `CMAKE_CXX_LINKER_LAUNCHER` to a shell script which dumps the linker command when linking the `link_info.cc` binary. Getting that linker command is the only reason that `link_info.cc` exists. I could have also picked one of the test targets as well but it felt a bit less hacky to just introduce something specific.

Once I had that linker command the idea was to parse it and then mirror it over to the Rust way of configuring linking. Unfortunately, all I end up with is a bunch of undefined symbols from libc++ and some macOS frameworks (which are kinda like system library things on macOS, I don't think Windows or Linux has an equivalent).

PR #140, opened 2024-07-22
Set local rpath lookups for libtiledb

davisp:
Adding $ORIGIN and @loader_path to the RPATH allows users to just place a copy of libtiledb.{so,dylib,dll} next to the binary instead of forcing it's presence in a specific system directory. This should allow for users to be able to play with the binaries created by rustc without having to setup a full build and install of libtiledb.

### DIFF
--- a/tiledb/api/build.rs
+++ b/tiledb/api/build.rs
@@ -1,5 +1,8 @@
 fn main() {
     let libdir = pkg_config::get_variable("tiledb", "libdir")
         .expect("Build-time TileDB library missing.");
-    println!("cargo:rustc-link-arg=-Wl,-rpath,{}", libdir);
+    println!(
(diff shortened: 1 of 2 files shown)


davisp:
@rroelke Can you double check this on linux? I tested locally by running a standard `cargo test`, and then moving my `/opt/tiledb/lib/libtiledb.dylib` to `target/debug/deps` and running it again. Its important that it doesn't exist wherever pkg-config found it is the main thing I'm looking to see working.

rroelke:
I don't have `libtiledb.dylib` but I tried with `libtiledb.so*`.

Notably `cargo clean` removes `target/debug/deps`.

Other than that I do believe it worked!

davisp:
Ah right, I forgot to translate from macOS to Linux there. But thanks for checking! Though I do think I've got a valid static build approach so this might not be needed. Though I'm also not sure its not unreasonable to just leave as is to give folks more options on where there `libtiledb.{dll,so,dylib}` lives.

PR #141, opened 2024-07-22
Wrap Array consolidation and vacuuming APIs

svilen-mihaylov-db:
…functions:

fn tiledb_array_consolidate
fn tiledb_array_consolidate_fragments
fn tiledb_array_consolidate_with_key
fn tiledb_array_upgrade_version
fn tiledb_array_vacuum

### DIFF
--- a/tiledb/api/src/array/mod.rs
+++ b/tiledb/api/src/array/mod.rs
@@ -35,6 +35,7 @@ pub use domain::{Builder as DomainBuilder, Domain, DomainData};
 pub use enumeration::{
     Builder as EnumerationBuilder, Enumeration, EnumerationData,
 };
+use ffi::tiledb_config_t;
 pub use fragment_info::{
     Builder as FragmentInfoBuilder, FragmentInfo, FragmentInfoList,
 };
@@ -196,6 +197,13 @@ impl ContextBound for Array {
     fn context(&self) -> Context {
         self.context.clone()
     }
+}
+
+fn unwrap_config_to_ptr(context: Option<&Config>) -> *mut tiledb_config_t {
+    context.map_or_else(
+        || std::ptr::null::<Config>() as *mut tiledb_config_t,
+        |ctx| ctx.capi(),
+    )
 }
 
 impl Array {
@@ -385,14 +393,18 @@ impl Array {
     pub fn vacuum<S>(
         ctx: &Context,
         array_uri: S,
-        config: &Config,
+        config: Option<&Config>,
     ) -> TileDBResult<()>
     where
         S: AsRef<str>,
     {
         let c_array_uri = cstring!(array_uri.as_ref());
         ctx.capi_call(|ctx| unsafe {
-            ffi::tiledb_array_vacuum(ctx, c_array_uri.as_ptr(), config.capi())
+            ffi::tiledb_array_vacuum(
+                ctx,
+                c_array_uri.as_ptr(),
+                unwrap_config_to_ptr(config),
+            )
         })?;
         Ok(())
     }
@@ -401,7 +413,7 @@ impl Array {
     pub fn upgrade_version<S>(
         ctx: &Context,
         array_uri: S,
-        config: &Config,
+        config: Option<&Config>,
     ) -> TileDBResult<()>
     where
         S: AsRef<str>,
@@ -411,7 +423,7 @@ impl Array {
             ffi::tiledb_array_upgrade_version(
                 ctx,
                 c_array_uri.as_ptr(),
-                config.capi(),
+                unwrap_config_to_ptr(config),
             )
         })?;
         Ok(())
@@ -422,7 +434,7 @@ impl Array {
     pub fn consolidate<S>(
         ctx: &Context,
         array_uri: S,
-        config: &Config,
+        config: Option<&Config>,
     ) -> TileDBResult<()>
     where
         S: AsRef<str>,
@@ -432,7 +444,7 @@ impl Array {
             ffi::tiledb_array_consolidate(
                 ctx,
                 c_array_uri.as_ptr(),
-                config.capi(),
+                unwrap_config_to_ptr(config),
             )
         })?;
         Ok(())
@@ -443,7 +455,7 @@ impl Array {
         ctx: &Context,
         array_uri: S,
         fragment_names: &[S],
-        config: &Config,
+        config: Option<&Config>,
     ) -> TileDBResult<()>
     where
         S: AsRef<str>,
@@ -466,7 +478,7 @@ impl Array {
                 c_array_uri.as_ptr(),
                 fragment_names_ptr.as_ptr(),
                 fragment_names_ptr.len() as u64,
-                config.capi(),
+                unwrap_config_to_ptr(config),
             )
         })?;
         Ok(())
@@ -1199,8 +1211,8 @@ pub mod tests {
         assert_eq!(4, count_fragments_fn()?);
 
         // Consolidate and Vacuum.
-        Array::consolidate(&ctx, &array_uri, &config).unwrap();
-        Array::vacuum(&ctx, &array_uri, &config).unwrap();
+        Array::consolidate(&ctx, &array_uri, Some(&config)).unwrap();
+        Array::vacuum(&ctx, &array_uri, Some(&config)).unwrap();
         // We have consolidated first two fragments.
         assert_eq!(3, count_fragments_fn()?);
 
@@ -1213,9 +1225,9 @@ pub mod tests {
             &ctx,
             array_uri.clone(),
             &fragment_names[1..3],
-            &config,
+            Some(&config),
         )?;
-        Array::vacuum(&ctx, &array_uri, &config).unwrap();
+        Array::vacuum(&ctx, &array_uri, Some(&config)).unwrap();
         assert_eq!(2, count_fragments_fn().unwrap());
 
         Ok(())
--- a/tiledb/api/src/array/mod.rs
+++ b/tiledb/api/src/array/mod.rs
@@ -405,8 +405,7 @@ impl Array {
                 c_array_uri.as_ptr(),
                 unwrap_config_to_ptr(config),
             )
-        })?;
-        Ok(())
+        })
     }
 
     /// Upgrades an array to the latest format version.
@@ -425,8 +424,7 @@ impl Array {
                 c_array_uri.as_ptr(),
                 unwrap_config_to_ptr(config),
             )
-        })?;
-        Ok(())
+        })
     }
 
     /// Depending on the consolidation mode in the config, consolidates either the fragment files,
@@ -446,8 +444,7 @@ impl Array {
                 c_array_uri.as_ptr(),
                 unwrap_config_to_ptr(config),
             )
-        })?;
-        Ok(())
+        })
     }
 
     /// Consolidates the given fragment URIs into a single fragment.
@@ -480,8 +477,7 @@ impl Array {
                 fragment_names_ptr.len() as u64,
                 unwrap_config_to_ptr(config),
             )
-        })?;
-        Ok(())
+        })
     }
 
     // Implements `dimension_nonempty_domain` for dimensions with CellValNum::Fixed
--- a/tiledb/api/src/array/mod.rs
+++ b/tiledb/api/src/array/mod.rs
@@ -26,6 +26,7 @@ pub mod enumeration;
 pub mod fragment_info;
 pub mod schema;
 
+use crate::config::Config;
 pub use attribute::{Attribute, AttributeData, Builder as AttributeBuilder};
 pub use dimension::{
     Builder as DimensionBuilder, Dimension, DimensionConstraints, DimensionData,
@@ -378,6 +379,97 @@ impl Array {
 
         let datatype = Datatype::try_from(c_datatype)?;
         Ok(Some(datatype))
+    }
+
+    /// Cleans up the array, such as consolidated fragments and array metadata.
+    pub fn vacuum<S>(
+        ctx: &Context,
+        array_uri: S,
+        config: &Config,
+    ) -> TileDBResult<()>
+    where
+        S: AsRef<str>,
+    {
+        let c_array_uri = cstring!(array_uri.as_ref());
+        ctx.capi_call(|ctx| unsafe {
+            ffi::tiledb_array_vacuum(ctx, c_array_uri.as_ptr(), config.capi())
+        })?;
+        Ok(())
+    }
+
+    /// Upgrades an array to the latest format version.
+    pub fn upgrade_version<S>(
+        ctx: &Context,
+        array_uri: S,
+        config: &Config,
+    ) -> TileDBResult<()>
+    where
+        S: AsRef<str>,
+    {
+        let c_array_uri = cstring!(array_uri.as_ref());
+        ctx.capi_call(|ctx| unsafe {
+            ffi::tiledb_array_upgrade_version(
+                ctx,
+                c_array_uri.as_ptr(),
+                config.capi(),
+            )
+        })?;
+        Ok(())
+    }
+
+    /// Depending on the consolidation mode in the config, consolidates either the fragment files,
+    /// fragment metadata files, or array metadata files into a single file.
+    pub fn consolidate<S>(
+        ctx: &Context,
+        array_uri: S,
+        config: &Config,
+    ) -> TileDBResult<()>
+    where
+        S: AsRef<str>,
+    {
+        let c_array_uri = cstring!(array_uri.as_ref());
+        ctx.capi_call(|ctx| unsafe {
+            ffi::tiledb_array_consolidate(
+                ctx,
+                c_array_uri.as_ptr(),
+                config.capi(),
+            )
+        })?;
+        Ok(())
+    }
+
+    /// Consolidates the given fragment URIs into a single fragment.
+    pub fn consolidate_fragments<S>(
+        ctx: &Context,
+        array_uri: S,
+        fragment_names: &[S],
+        config: &Config,
+    ) -> TileDBResult<()>
+    where
+        S: AsRef<str>,
+    {
+        let c_array_uri = cstring!(array_uri.as_ref());
+
+        // This array has to outlive the API call below.
+        let fragment_names_cstr = fragment_names
+            .iter()
+            .map(|fragment_name| Ok(cstring!(fragment_name.as_ref())))
+            .collect::<TileDBResult<Vec<_>>>()?;
+        let fragment_names_ptr = fragment_names_cstr
+            .iter()
+            .map(|fragment_name| fragment_name.as_ptr())
+            .collect::<Vec<_>>();
+
+        ctx.capi_call(|ctx| unsafe {
+            ffi::tiledb_array_consolidate_fragments(
+                ctx,
+                c_array_uri.as_ptr(),
+                fragment_names_ptr.as_ptr(),
+                fragment_names_ptr.len() as u64,
+                config.capi(),
+            )
+        })?;
+        Ok(())
     }
 
     // Implements `dimension_nonempty_domain` for dimensions with CellValNum::Fixed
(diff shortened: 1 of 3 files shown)


tiledb/api/src/array/mod.rs
```diff
@@ -380,6 +382,113 @@ impl Array {
         Ok(Some(datatype))
     }
 
+    pub fn vacuum<S>(
```
svilen-mihaylov-db, line 385: New apis for vacuuming and consolidation

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
 
         test_uri.close().map_err(|e| Error::Other(e.to_string()))
     }
+
+    fn create_simple_dense(
```
svilen-mihaylov-db, line 1134: creating dense vector

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
...
+            .build()?;
+
+        let array_uri = test_uri
+            .with_path("quickstart_dense")
+            .map_err(|e| Error::Other(e.to_string()))?;
+        Array::create(ctx, &array_uri, schema)?;
+        Ok(array_uri)
+    }
+
+    fn write_dense_vector_4_fragments(
```
svilen-mihaylov-db, line 1163: writing four fragments

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
...
+
+        let count_fragment_fn = || -> TileDBResult<u32> {
+            FragmentInfoBuilder::new(&ctx, array_uri.clone())?
+                .build()?
+                .unconsolidated_metadata_num()
+        };
+        assert_eq!(4, count_fragment_fn().unwrap());
+
+        // Consolidate and Vacuum
+        Array::consolidate(&ctx, &array_uri, &config).unwrap();
```
svilen-mihaylov-db, line 1224: consolidate and vacuum

tiledb/api/src/array/mod.rs
```diff
@@ -380,6 +382,113 @@ impl Array {
...
+            ffi::tiledb_array_consolidate(
+                ctx,
+                c_array_uri.as_ptr(),
+                config.capi(),
+            )
+        })?;
+        Ok(())
+    }
+
+    pub fn consolidate_fragments<S>(
```
svilen-mihaylov-db, line 438: TODO: test consolidate fragments and consolidate fragments with key.

tiledb/api/src/query/subarray.rs
```diff
@@ -359,6 +359,38 @@ where
...
     }
 
+    pub fn set_subarray<IntoRange: Into<Range>>(
+        self,
+        range: IntoRange,
+    ) -> TileDBResult<Self> {
+        let c_subarray = *self.raw;
+
+        let range = range.into();
+        // TODO: should we be checking dimensionality here?
```
svilen-mihaylov-db, line 369: Have some questions here...

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: Also perhaps split those for a future pr?

tiledb/api/src/array/mod.rs
```diff
@@ -814,7 +925,7 @@ pub mod tests {
                 .build()
         };
 
-        let s: Schema = SchemaBuilder::new(context, ArrayType::Sparse, d)
+        let s: Schema = SchemaBuilder::new(context, ArrayType::Dense, d)
```
svilen-mihaylov-db, line 928: This seemed unintentional to me. The function is named "create_quickstart_dense".

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
...
+                .build()?
+                .unconsolidated_metadata_num()
+        };
+        assert_eq!(4, count_fragment_fn().unwrap());
+
+        // Consolidate and Vacuum
+        Array::consolidate(&ctx, &array_uri, &config).unwrap();
+        Array::vacuum(&ctx, &array_uri, &config).unwrap();
+
+        assert_eq!(3, count_fragment_fn().unwrap());
```
svilen-mihaylov-db, line 1227: Down to 3 segments from 4.

tiledb/api/src/array/mod.rs
```diff
@@ -380,6 +382,113 @@ impl Array {
...
+        S: AsRef<str>,
+    {
+        let c_array_uri = cstring!(array_uri.as_ref());
+        ctx.capi_call(|ctx| unsafe {
+            ffi::tiledb_array_vacuum(ctx, c_array_uri.as_ptr(), config.capi())
+        })?;
+        Ok(())
+    }
+
+    pub fn upgrade_version<S>(
```
svilen-mihaylov-db, line 400: I was considering using a lambda to simplify the wrapping if possible, since the code is very repetitive. The lambda would be either tiledb_array_vaccum, tiledb_upgrade_version, or another call with three arguments {context, ari, config}

tiledb/api/src/query/subarray.rs
```diff
@@ -359,6 +359,38 @@ where
...
+                        // TODO: is this correct?
+                        ffi::tiledb_subarray_set_subarray(
+                            ctx,
+                            c_subarray,
+                            start.as_ptr() as *const std::ffi::c_void,
+                        )
+                    })?;
+                })
+            }
+            Range::Multi(_) => unreachable!("Unreachable (TODO)"),
```
rroelke, line 387: Rust has `unimplemented!()` and `todo!()` as well which are quite nice.

tiledb/api/src/query/subarray.rs
rroelke: I'm basing the following interpretation off of the example `multi_attribute.c`:
```
  // Slice only rows 1, 2 and cols 2, 3, 4
  tiledb_subarray_t* subarray;
  tiledb_subarray_alloc(ctx, array, &subarray);
  int subarray_v[] = {1, 2, 2, 4};
  tiledb_subarray_set_subarray(ctx, subarray, subarray_v);
```

I think what the C function `tiledb_subarray_set_subarray` does is
(a) assume that all dimensions are the same type (which for a dense array they are)
(b) assume that the user input contains two coordinates per dimension
(c) for dimension `d`, set the lower bound of the subarray for that dimension to be the input array index `2 * d` and the upper bound to be array index `2 * d + 1`


So, what you have here works for one dimension since the input is one range, but I imagine for arrays with more dimensions there will be undefined behavior.  Hence to directly answer the question - yes, this definitely should check that the number of input ranges matches the dimensionality!

If you change the input to be a `Vec<Range>` or `Iterator<Item = Range>` or whatever, then you can do a bunch of work to buffer the coordinates and then call `ffi::tiledb_subarray_set_subarray`... but you could also do something like `.enumerate().fold(self, |subarray, (dimension, range)| self.add_range(dimension, range))` and avoid type checks and buffering and etc.

tiledb/api/src/array/mod.rs
rroelke: Either way is fine with me.

tiledb/api/src/array/mod.rs
rroelke: It's a minor style choice but I feel a little bit like these should be module-level functions rather than part of the `impl`, @davisp do you have a preference?

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
...
+    ) -> TileDBResult<()> {
+        // Subarray boundaries.
+        let boundaries = [0, 200, 250, 310, 410];
+
+        for i in 0..4 {
+            // Prepare cell buffer.
+            let low_bound = boundaries[i];
+            let high_bound = boundaries[i + 1];
+
+            let mut data: Vec<i32> = vec![];
```
rroelke, line 1176: `(low_bound.. high_bound).collect::<Vec<i32>>()` is more idiomatic

(and, not that it matters here but might some other place, will reserve space at the beginning to do one allocation)

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
...
+            for j in low_bound..high_bound {
+                data.push(j);
+            }
+
+            let opener = ArrayOpener::new(ctx, array_uri, Mode::Write)?
+                .end_timestamp(timestamp + i as u64 + 1)?;
```
rroelke, line 1182: I think the timestamps are epoch times, not versions, so `timestamp + i` is probably not going to be what you want.

For writing an array I'm not sure what the open timestamp actually does though.  What is your intent?

tiledb/api/src/array/mod.rs
```diff
@@ -1019,4 +1130,102 @@ pub mod tests {
...
+        config.set("sm.consolidation.steps", "1").unwrap();
+        config.set("sm.consolidation.step_min_frags", "2").unwrap();
+        config.set("sm.consolidation.step_max_frags", "2").unwrap();
+        config
+            .set("sm.consolidation.step_size_ratio", "0.0")
+            .unwrap();
+        config.set("sm.consolidation.buffer_size", "10000").unwrap();
+
+        let count_fragment_fn = || -> TileDBResult<u32> {
+            FragmentInfoBuilder::new(&ctx, array_uri.clone())?
```
rroelke, line 1217: This is a little odd. I know it's the way to do it with the current API but I feel like we should have `FragmentInfoList::open` or something.

tiledb/api/src/array/mod.rs
rroelke: I'd recommend adding them to this one just so we don't forget.  Stories for "add tests to thing" tend to languish, in my experience.

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: Everything seems to pass with the change. I can introduce a parameter to optionally create dense or sparse. Does that work for you?

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: I am following what ever this test is doing: https://github.com/TileDB-Inc/TileDB/blob/2a7935e46575e6d1dfaed800ade44fdef15cfc0d/test/src/unit-capi-consolidation.cc#L5091

To your specific question those are the lines in the test which adds the timestamps: 
https://github.com/TileDB-Inc/TileDB/blob/2a7935e46575e6d1dfaed800ade44fdef15cfc0d/test/src/unit-capi-consolidation.cc#L717
https://github.com/TileDB-Inc/TileDB/blob/2a7935e46575e6d1dfaed800ade44fdef15cfc0d/test/src/unit-capi-consolidation.cc#L765
https://github.com/TileDB-Inc/TileDB/blob/2a7935e46575e6d1dfaed800ade44fdef15cfc0d/test/src/unit-capi-consolidation.cc#L808
https://github.com/TileDB-Inc/TileDB/blob/2a7935e46575e6d1dfaed800ade44fdef15cfc0d/test/src/unit-capi-consolidation.cc#L852

I have simply refactored in a loop.

tiledb/api/src/array/mod.rs
```diff
@@ -380,6 +382,113 @@ impl Array {
...
+    ) -> TileDBResult<()>
+    where
+        S: AsRef<str>,
+    {
+        let c_array_uri = cstring!(array_uri.as_ref());
+
+        let v = fragment_names
+            .iter()
+            .map(|fragment_name| Ok(cstring!(fragment_name.as_ref()).as_ptr()))
+            .map(|v| v.unwrap())
```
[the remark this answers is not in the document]
svilen-mihaylov-db: The issue here is that on line 459 below I need to obtain a pointer to the vector's storage, so I need to eventually unwrap the result of cstring it seems to me. How else should I do it?

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: Thanks for the suggestion. I can also add that I can let the data be immutable.

tiledb/api/src/query/subarray.rs
svilen-mihaylov-db: I am following the example from here: https://github.com/TileDB-Inc/tiledb-rs/blob/bcf910a94b32202bf8e560bf3beb424ed8012677/tiledb/api/src/query/subarray.rs#L264

This is more of a question. Should I be worrying about Range::Multi and Range::Var at this point. I'm not clear about the intent behind Single, Multi and Var.

tiledb/api/src/array/mod.rs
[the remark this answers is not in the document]
rroelke: If you do `collect::<TileDBResult<Vec<_>>>()?` with the question mark operator, then it will return `Err` if the result is an error and otherwise the expression evaluates to `Vec<_>`

tiledb/api/src/array/mod.rs
rroelke: If everything passes I don't have a preference. As long as the function name isn't lying :)

tiledb/api/src/array/mod.rs
rroelke: Oh, I agree that this is the best that can be done with the current way of the world, I just thing we should improve the way of the world at some point.

tiledb/api/src/array/mod.rs
rroelke: Hm, I think that test is most likely wrong.  Maybe at some point in the past it was a version but then changed to a timestamp?

I did check one of our other tests which uses the timestamp ranges, and the data on the fragments is definitely epoch time. Or some other choice of many-digit number :)

Or possibly it's benign.  As I alluded I have no idea what the timestamp range does for a *write* - I'm guessing if you just remove this it will do the same thing but be less confusing.

tiledb/api/src/query/subarray.rs
rroelke: I don't think so.
`Range::Multi` AFAIK is not used at all in practice, maybe it will be someday but isn't there yet.
`Range::Var` can show up on string-typed dimensions for sparse arrays, but for those one would have to use a different function to set the range.


Aside:

Attributes and dimensions both have a "cell val num" which is the number of values of the type that occur per cell in the array.
- a value of 1 is kind of what you would expect
- a fixed value N of greater than 1 means that exactly N values appear per cell.  This is not common, unclear if it is ever used at all.  If it is, surely it is only small numbers.  And it definitely never appears on a dimension.
- `CellValNum::Var` is what it looks like in `tiledb-rs`, in core this is the constant `u32::MAX`, but either way this value means that the number of values per cell is variable.  To query these in C you need an additional offsets column - in `tiledb-rs` it can be done that way but there are also iterators to abstract it away.

Dimensions of dense arrays always have cell val num 1.  Dimensions of sparse arrays do too, *except* `StringAscii` typed dimensions can have `CellValNum::Var`.

tiledb/api/src/array/mod.rs
davisp: I don't have a preference in either direction. The only thought I have to push one way or the other is to check the C++ API which has them as static methods on Array as well:

https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/cpp_api/array.h#L588-L594

tiledb/api/src/array/mod.rs
davisp: I can speak more to it on the call this afternoon, but the basic idea is this: Every write happens at a certain time. This time is encoded into the fragment URIs in the form of `__t1_t2_uuid_v` [(Code here)](https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/storage_format/uri/generate_uri.cc#L41-L65) These tests that are operating at the fragment level will often specify their timestamps manually so that various assertions can be made about how fragment handling works.

However, if you see tests that are doing this but not making specific fragment assertions it can also be a historical artifact due to timing bugs. The timestamps in our filenames are millisecond resolution, which meant that any two fragments written in the same millisecond would be ordered randomly based on the UUID component. This has since been fixed so that any two writes in the same OS process will now be correctly ordered regardless of whether they happen in the same millisecond.

A quick skim of these tests suggests this is probably a reflection of the historical issues around timing since we're not attempting to do fancier consolidation stuff. Though I'm pretty sure there are tests that do that somewhere that would still need them (i.e., if this test were consolidating the two middle fragments we'd need to manually specify timestamps so that they'd be selectable).

davisp (commented):
Overall looks good to me.

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: If I understand correctly the inclination is then to leave them as impl (static)?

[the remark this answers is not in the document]
svilen-mihaylov-db: Do you suggest I start with the c api documentation?

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: Ok. I will leave as above (use Dense)

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: I'm not following what exactly the proposal is here? What should I be correcting wrt above implementation?

tiledb/api/src/query/subarray.rs
svilen-mihaylov-db: Thanks for explaining, will update based on comments above.

svilen-mihaylov-db: To be clear, do you propose both be unimplemented?

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: Per standup discussion will leave as above.

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: per discussion, will leave as is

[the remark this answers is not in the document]
svilen-mihaylov-db: Will add comments based on c api.

tiledb/api/src/query/subarray.rs
svilen-mihaylov-db: Is it correct to say you are suggesting to shim this into a call to .add_range()?

rroelke: Yep.  I think we should add `tiledb_subarray_set_subarray` to the ignored function list.

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: done

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: left as is

tiledb/api/src/array/mod.rs
```diff
@@ -380,6 +381,97 @@ impl Array {
...
+            ffi::tiledb_array_vacuum(ctx, c_array_uri.as_ptr(), config.capi())
+        })?;
+        Ok(())
+    }
+
+    /// Upgrades an array to the latest format version.
+    pub fn upgrade_version<S>(
+        ctx: &Context,
+        array_uri: S,
+        config: &Config,
```
rroelke, line 404: Hm, maybe this should be `Option` as it sounds from the C API description that it is not required.

tiledb/api/src/array/mod.rs
```diff
@@ -380,6 +381,97 @@ impl Array {
...
+        })?;
+        Ok(())
+    }
+
+    /// Consolidates the given fragment URIs into a single fragment.
+    pub fn consolidate_fragments<S>(
+        ctx: &Context,
+        array_uri: S,
+        fragment_names: &[S],
+        config: &Config,
```
rroelke, line 446: As above, this looks optional, so `Option<&Config>` might be better

rroelke (approved):
Thanks for making the changes, looks good!  I left a note about using `Option<&Config>` for the new functions, it sounds like from the C API these are optional parameters and we should reflect that. But that's a simple change so I don't need to review it again.

tiledb/api/src/array/mod.rs
svilen-mihaylov-db: I've checked the c api and looks like it accepts null, so I will update. Same as below.

PR #142, opened 2024-07-23
Various API additions and one tweak

rroelke:
Some small usability additions as well as a name change: `Domain::ndim` to `Domain::num_dimensions`, to match the style of our other `num_thing` methods.

The underlying functions are already tested, except for the addition to the test code `Cells` but I plan to use that in a branch I'm working on.

### DIFF
--- a/tiledb/api/src/array/attribute/arrow.rs
+++ b/tiledb/api/src/array/attribute/arrow.rs
@@ -15,6 +15,20 @@ use crate::filter::arrow::FilterMetadata;
 use crate::filter::FilterListBuilder;
 use crate::{physical_type_go, Context, Datatype, Result as TileDBResult};
 
+// additional methods with arrow features
+impl Attribute {
+    pub fn to_arrow(&self) -> TileDBResult<FieldToArrowResult> {
+        crate::array::attribute::arrow::to_arrow(self)
+    }
+
+    pub fn from_arrow(
+        context: &Context,
+        field: &arrow::datatypes::Field,
+    ) -> TileDBResult<AttributeFromArrowResult> {
+        crate::array::attribute::arrow::from_arrow(context, field)
+    }
+}
+
 /// Encapsulates TileDB Attribute fill value data for storage in Arrow field metadata
 #[derive(Deserialize, Serialize)]
 pub struct FillValueMetadata {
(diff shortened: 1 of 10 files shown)


tiledb/api/src/array/attribute/arrow.rs
```diff
@@ -15,6 +15,20 @@ use crate::filter::arrow::FilterMetadata;
 use crate::filter::FilterListBuilder;
 use crate::{physical_type_go, Context, Datatype, Result as TileDBResult};
 
+// additional methods with arrow features
```
svilen-mihaylov-db, line 18: nit: I'm not sure if we have guidelines around it, but for function or module-level comments I have a preference for block comments.

tiledb/api/src/query/strategy.rs
```diff
@@ -1026,6 +1026,36 @@ impl Cells {
 
         self.filter(&preserve)
     }
+
+    /// Returns a copy of `self` with only the fields in `fields`,
+    /// or `None` if not all the requested fields are present.
+    pub fn projection(&self, fields: &[&str]) -> Option<Cells> {
```
svilen-mihaylov-db, line 1032: Can you point me to usages of those two new functions?

tiledb/api/src/array/attribute/arrow.rs
davisp: Guidelines for idiomatic rust are to use the `//` on each line: https://doc.rust-lang.org/book/ch03-04-comments.html

davisp (approved):
+1

tiledb/api/src/query/strategy.rs
rroelke: This is maybe an example of where splitting the repositories `tiledb-rs` and `tiledb-tables` is sort of inconvenient.  At some point I was using this function in my development branch of `tables`, but now I've realized I no longer need it, and yet here it remains.

Note that the entirety of this file (and anything you see in any `strategy.rs`) amounts to testing code.

As you may see from the other commit I just pushed to this branch, sometimes things that "obviously work" don't... so I'll add some tests for this function

svilen-mihaylov-db: Are you saying that you plan to add coverage in a follow up pr?

rroelke: I pushed a commit to this branch

https://github.com/TileDB-Inc/tiledb-rs/pull/142/commits/e489647e8fc72798ef67516beb4d7878c30687c3

svilen-mihaylov-db: I do not entirely understand this, I do not want to hold this cr.

PR #144, opened 2024-07-29
Set local rpath lookups for libtiledb

davisp:
Adding $ORIGIN and @loader_path to the RPATH allows users to just place a copy of libtiledb.{so,dylib,dll} next to the binary instead of forcing it's presence in a specific system directory. This should allow for users to be able to play with the binaries created by rustc without having to setup a full build and install of libtiledb.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1332,6 +1332,7 @@ name = "tiledb-sys"
 version = "0.1.0"
 dependencies = [
  "pkg-config",
+ "tiledb-utils",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


davisp:
This replaces #140. The only notable differences are that the RPATH print is moved to a utility function in `tiledb-utils` so that it can be reused where needed along with a change to the order so that relative libraries are searched first to hopefully avoid any issues with users that have a conda installed version of libtiledb that isn't compatible with the conda nightly channel.

rroelke (approved):

PR #145, opened 2024-08-05
chore: [sc-52402] [rs] impl Display for QueryConditionExpr

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/52402
Blocks https://app.shortcut.com/tiledb-inc/story/52316/tables-cli-explain-query-panics-with-not-implemented

> For tables we would like details about the ArrayScan operator to appear in EXPLAIN plan output. The verbose EXPLAIN output should include as many details as possible about the scan, including subarrays and query conditions.

To do the above work for tables we must first `impl Display for QueryConditionExpr` in the `api` crate.

### DIFF
--- a/tiledb/api/src/query/conditions.rs
+++ b/tiledb/api/src/query/conditions.rs
@@ -880,7 +880,4 @@ mod tests {
             qc_tree.to_string()
         );
     }
-
-    #[test]
-    fn display_ambiguity() {}
 }
--- a/tiledb/api/src/query/conditions.rs
+++ b/tiledb/api/src/query/conditions.rs
@@ -1,3 +1,4 @@
+use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::ops::{BitAnd, BitOr, Deref, Not};
 
 use anyhow::anyhow;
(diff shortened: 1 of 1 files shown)


tiledb/api/src/query/conditions.rs
```diff
@@ -714,4 +848,39 @@ mod tests {
...
     }
+
+    #[test]
+    fn display() {
+        let qc_cmp = QC::field("field").lt(5);
+        assert_eq!("field < 5", qc_cmp.to_string());
+
+        let qc_setmemb =
+            QC::field("field").is_in(["one", "two", "three"].as_slice());
+        assert_eq!("field IN (one, two, three)", qc_setmemb.to_string());
```
rroelke, line 859: Oops, need to escape these.

svilen-mihaylov-db (approved):

tiledb/api/src/query/conditions.rs
```diff
@@ -714,4 +848,39 @@ mod tests {
...
+            & (!atom.clone() | !(atom.clone() & atom.clone()));
+
+        assert_eq!(
+            "((x < 5) OR (x < 5)) AND ((NOT (x < 5)) OR (NOT ((x < 5) AND (x < 5))))",
+            qc_tree.to_string()
+        );
+    }
+
+    #[test]
+    fn display_ambiguity() {}
```
davisp, line 885: Not sure I understand what this test is testing.

davisp (approved):
+1

tiledb/api/src/query/conditions.rs
rroelke: That seems consistent with the name of the test!


To be serious, I started putting the nested case here but then decided to just have the one test function

tiledb/api/src/query/conditions.rs
rroelke: I am so, so thankful that `str::escape_default` is a function that exists.

PR #146, opened 2024-08-06
Remove cfg(test) for Datatype::iter

rroelke:
This function is conceivably useful for test code in crates using `tiledb-rs`.  For example, `tables`.


davisp (approved):
+1

davisp:
Hah! Talk about apropos! I just realized I could really use a list of all datatypes for testing and then... oh right, we just made that happen this morning!

PR #154, opened 2024-08-27
Pull out testing / generally useful code from the aggregate API PR

rroelke:
The commits in this PR are essentially unrelated to each other, the common theme merely being that these are all changes I made while working on #113 .  I decided to pull them out so aggregates will take up a greater percentage of the lines of code in their introductory PR.

Anyway, looking at this commit-by-commit is not a bad idea; notes on each below.

[Add asserts to attribute_default about nullability](https://github.com/TileDB-Inc/tiledb-rs/commit/70fb2cfccb2636f46317b242178df5fd1e853e71) - we make assumptions about default core library behavior in a few places, so let's make sure if it ever changes we get a concise alert.

[strat_dimension, strat_attribute, strat_field](https://github.com/TileDB-Inc/tiledb-rs/commit/04a381ae9c2d693296d75e57ff5e0a5ac75a88eb) - proptest strategies for picking random schema fields

[impl TryFrom<FieldData> for Vec<_>](https://github.com/TileDB-Inc/tiledb-rs/commit/07617231cf1b33a14740d2188d62c47e6d83a7b2) - for making generic programming with the test data structures a bit easier

[Error::physical_type_mismatch](https://github.com/TileDB-Inc/tiledb-rs/commit/2e950f6e7a0a9b878357ecbf55df958432770a0d) - we have a `TypeMismatch` variant already for matching an incorrect physical type with a `Datatype`, but we don't always have a `Datatype`.

[API crate add module for internal test utilities](https://github.com/TileDB-Inc/tiledb-rs/commit/532449026fab518e2ad6a011804f6b31ab0d6516) - this is the largest change here, our test code has tons of common imports and boilerplate.  We have a bunch of test utils crates, but we need to have something inside the `api` crate in order to avoid circular dependencies.  The `sparse_all` schema is one I will use soon in the aggregates PR and the `TestArray` struct was helpful for avoiding large duplicated code paragraphs that ultimately distract from the actual goal of the tests.

### DIFF
--- a/tiledb/api/src/error.rs
+++ b/tiledb/api/src/error.rs
@@ -200,6 +200,7 @@ pub enum Error {
 }
 
 impl Error {
+    #[cfg(test)]
     pub(crate) fn physical_type_mismatch<T, U>() -> Self {
         Self::Datatype(DatatypeErrorKind::PhysicalTypeMismatch {
             requested_type: std::any::type_name::<T>().to_owned(),
--- a/tiledb/api/src/error.rs
+++ b/tiledb/api/src/error.rs
@@ -33,6 +33,10 @@ pub enum DatatypeErrorKind {
         user_type: String,
         tiledb_type: Datatype,
     },
+    PhysicalTypeMismatch {
+        requested_type: String,
+        actual_type: String,
+    },
     UnexpectedCellStructure {
         context: Option<String>,
         found: CellValNum,
@@ -62,6 +66,16 @@ impl Display for DatatypeErrorKind {
                     f,
                     "Type mismatch: requested {}, but found {}",
                     user_type, tiledb_type
+                )
+            }
+            DatatypeErrorKind::PhysicalTypeMismatch {
+                requested_type,
+                actual_type,
+            } => {
+                write!(
+                    f,
+                    "Physical type mismatch: requested {}, but found {}",
+                    requested_type, actual_type
                 )
             }
             DatatypeErrorKind::UnexpectedCellStructure {
@@ -183,6 +197,15 @@ pub enum Error {
     /// Any error which cannot be categorized as any of the above
     #[error("{0}")]
     Other(String),
+}
+
+impl Error {
+    pub(crate) fn physical_type_mismatch<T, U>() -> Self {
+        Self::Datatype(DatatypeErrorKind::PhysicalTypeMismatch {
+            requested_type: std::any::type_name::<T>().to_owned(),
+            actual_type: std::any::type_name::<U>().to_owned(),
+        })
+    }
 }
 
 impl From<RawError> for Error {
--- a/tiledb/api/src/error.rs
+++ b/tiledb/api/src/error.rs
@@ -200,7 +200,6 @@ pub enum Error {
 }
 
 impl Error {
-    #[cfg(test)]
     pub(crate) fn physical_type_mismatch<T, U>() -> Self {
         Self::Datatype(DatatypeErrorKind::PhysicalTypeMismatch {
             requested_type: std::any::type_name::<T>().to_owned(),
--- a/tiledb/api/src/tests/examples/sparse_all.rs
+++ b/tiledb/api/src/tests/examples/sparse_all.rs
@@ -1,0 +1,148 @@
+use std::rc::Rc;
+
+use crate::array::{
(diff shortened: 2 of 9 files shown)


tiledb/api/src/tests/mod.rs
```diff
@@ -0,0 +1,29 @@
+pub mod examples;
+
+pub mod prelude {
+    pub mod array {
```
davisp, line 4: This prelude pattern seems odd to me. Googling I'm not really finding any specific "this is an idiomatic prelude" type guidance so there's nothing I can point to specifically except to note that every prelude I've seen before was intended to be `use thing::prelude::*` so having `use thing::prelude::other_thing::*` seems a bit awkward. `use thing::other_thing::prelude::*` or just removing the arra/query/strategy sub-modules altogether seems better.

But given I can't find a single example "most folks do it this way" I'll not request anything specific and will just consider this food for thought.

davisp (approved):
Other than the random thought on how to structure that prelude, this is all straightforward.

+1

tiledb/api/src/tests/mod.rs
rroelke: Yeah that's fair. I figured it would be a start, and it's internal anyway due to being inside `tests` module.

Having `prelude::*` be the trailing part of the path sounds nice so I'll re-organize this a bit.

tiledb/api/src/tests/examples/sparse_all.rs
```diff
@@ -0,0 +1,148 @@
...
+pub type FnAcceptAttribute =
+    dyn Fn(&Parameters, Datatype, CellValNum, bool) -> bool;
+
+/// Configures construction of the `sparse_all` schema.
+#[derive(Clone)]
+pub struct Parameters {
+    /// Function which determines whether to add a dimension to the schema.
+    ///
+    /// By default, all types are added as dimensions except `Datatype::StringAscii`.
+    pub fn_accept_dimension: Rc<FnAcceptDimension>,
```
svilen-mihaylov-db, line 19: Do we expect to modify those? Cant we use a Box type instead of reference counting?

rroelke: `Rc` is necessary for `#[derive(Clone)]`.
If we had `Box<dyn Thing>` then we can't clone it without requiring `Thing` to be `Clone`.
`Rc` seemed like a better choice.

svilen-mihaylov-db: Is cloned essential here though? Can you explain why we need to clone?

rroelke: Cloning is not essential but it can be useful for a partial initialization, e.g.
```
let whatever = Parameters {
    fn_accept_dimension: Rc::new(fn_accept_uint64_only),
    ..Default::default()
};
```

rroelke: Oops, `Default::default()` is a poor example, because we don't actually need `Clone` for that.  Whatever the expression is, gets moved from; so if we are trying to mostly copy an existing set of parameters but change one field, then we want to have `Clone` for that to be convenient.
```
let whatever = Parameters {
    fn_accept_dimension: Rc::new(fn_accept_uint64_only),
    ..existing_parameters.clone()
};
```

This is a pattern we see in some of the proptest code, e.g.
```
                let domain_requirements = Rc::new(DomainRequirements {
                    array_type: Some(array_type),
                    ..domain_requirements.as_ref().clone()
                });
```

svilen-mihaylov-db: Ok thanks for explaining. Resolving here.

tiledb/api/src/error.rs
```diff
@@ -185,6 +199,15 @@ pub enum Error {
     Other(String),
 }
 
+impl Error {
+    pub(crate) fn physical_type_mismatch<T, U>() -> Self {
```
svilen-mihaylov-db, line 203: #[cfg(test)]

rroelke: It is also used for the proptest-strategies feature, i.e. `cargo check --features proptest-strategies` fails to compile without it.

We can do `#[cfg(any(test, feature = "proptest-strategies"))]` as we do in other places, but I do anticipate that we will find a use for it somewhere that does not require this configuration.

PR #155, opened 2024-09-04
DimensionData infers CellValNum from DimensionConstraints

rroelke:
The `DimensionData::cell_val_num` field never tells us anything we can't infer from `DimensionData::constraints`.  Replacing `DimensionData::cell_val_num` with a function which does not return `Option` makes it a little more user-friendly.


davisp (approved):
+1

PR #157, opened 2024-09-05
Configure array schema proptest constants using environment; and some incidentals

rroelke:
A goal we haven't focused on much yet is configuring our nightly runs to generate more complicated inputs and test cases in our proptests.

This pull request implements one step towards this goal - adding environment variables for overriding the default values used in our array schema tests.

Toying with some of these environment variables exposed some bugs which are also incidentally addressed here, such as SC-51250

We also `impl Arbitrary for EnumerationData` along the way to this change.

### DIFF
--- a/tiledb/api/src/query/strategy.rs
+++ b/tiledb/api/src/query/strategy.rs
@@ -1561,24 +1561,47 @@ pub struct CellsParameters {
     pub schema: Option<CellsStrategySchema>,
     pub min_records: usize,
     pub max_records: usize,
-    pub value_min_var_size: usize,
-    pub value_max_var_size: usize,
+    pub cell_min_var_size: usize,
+    pub cell_max_var_size: usize,
+}
+
+impl CellsParameters {
+    pub fn min_records_default() -> usize {
+        const CELLS_MIN_RECORDS: usize = 0;
+        let env_min = "TILEDB_STRATEGY_CELLS_PARAMETERS_NUM_RECORDS_MIN";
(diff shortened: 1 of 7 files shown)


tiledb/api/src/query/strategy.rs
```diff
@@ -1598,22 +1624,7 @@ impl CellsStrategy {
     fn nrecords_limit(&self) -> Option<usize> {
         if let Some(schema) = self.schema.array_schema() {
             if !schema.allow_duplicates.unwrap_or(true) {
-                /*
```
rroelke, line 1601: 1. The iteration that is removed is essentially the same as the body of `schema.domain.num_cells()`
2. the issue that this comment mentions is resolved by code that has been written after this comment (in `CellStrategySchema::new_field_tree` near the comment `dimension coordinates must be unique, generate them first`)

davisp (approved):
This all looks good so I'm approving.

Though I will put a bee in your bonnet and ask if it might be a bit cleaner to have a single module/API where all of the "read configuration from environment" code is localized so that all of our settings are in a single file for future maintainability/discovery. Given this is all CI/test code I'm not going to request it since its only ever going to affect people actually doing development on the repo and none of our possible downstream users of tiledb-rs.

rroelke:
> Though I will put a bee in your bonnet and ask if it might be a bit cleaner to have a single module/API where all of the "read configuration from environment" code is localized so that all of our settings are in a single file for future maintainability/discovery.

Absolutely on board. Will merge this and then either do it or file a story.

PR #159, opened 2024-09-09
ignore domain_shrinking test

rroelke:
I'm not sure whether something changed recently, but I had a look at this test (whose goal is to ensure that shrinking a domain reduces the number of dimensions) and the way that it runs appears to spend most of the time shrinking the *name* of the domain.

I don't think that shrinking the name is something we should never do, but it certainly isn't worth waiting a minute for each time we run this test.

I've chosen to ignore this test rather than do something else because I think we should revisit the domain shrinking with a custom implementation and test that.  Removing the test also seems acceptable under this view.

### DIFF
--- a/tiledb/api/src/array/domain/strategy.rs
+++ b/tiledb/api/src/array/domain/strategy.rs
@@ -155,6 +155,7 @@ mod tests {
         });
     }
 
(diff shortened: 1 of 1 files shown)


davisp (approved):
+1

PR #160, opened 2024-09-10
impl Eq and Hash for SubarrayData and QueryConditionExpr

rroelke:
The purpose of this PR is to enable deriving Eq and Hash for ArrayScanFilterPushDown in the tables repo.

We need to do *that* so that we can put ArrayScanFilterPushDown into custom logical plan nodes, which must impl Eq and Hash.

We need to write custom logical plan nodes which embed ArrayScanFilterPushDown in order to push down aggregates (this is not exactly 100% true but it does lead to a nice implementation).

The only philosophical issue at hand is whether it is correct to do this because of the floating-point variants.  These are handled using our `BitsEq` and `BitsHash` traits respectively, which I claim accounts for correctness.  So, philosophically: floating point numbers do not impl Eq or Hash due to NaN.  I don't anticipate that we particularly care about propagating `NaN != NaN` as we might see it in predicates, ranges, or query conditions; we sacrifice some mathematical purity for significant improvements to usability.

### DIFF
--- a/tiledb/api/src/datatype/physical.rs
+++ b/tiledb/api/src/datatype/physical.rs
@@ -343,31 +343,3 @@ where
         self.0.bits_cmp(&other.0)
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use std::hash::DefaultHasher;
-
-    use proptest::prelude::*;
-
-    use super::*;
-
-    #[test]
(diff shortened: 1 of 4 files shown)


davisp (approved):
+1

PR #161, opened 2024-09-11
Finish basic support for enumerations

davisp:
This finishes out the basic support for adding enumeration support to arrays and schemas.

### DIFF
--- a/tiledb/api/src/array/attribute/mod.rs
+++ b/tiledb/api/src/array/attribute/mod.rs
@@ -17,6 +17,7 @@ use crate::datatype::PhysicalType;
 use crate::error::{DatatypeErrorKind, Error};
 use crate::filter::list::{FilterList, FilterListData, RawFilterList};
 use crate::physical_type_go;
+use crate::string::{RawTDBString, TDBString};
 use crate::{Datatype, Factory, Result as TileDBResult};
 
 pub(crate) enum RawAttribute {
@@ -195,6 +196,25 @@ impl Attribute {
         let is_valid = c_validity != 0;
         Ok((F::from_raw(slice)?, is_valid))
     }
+
+    /// Get the enumeration name
+    pub fn enumeration_name(&self) -> TileDBResult<Option<String>> {
+        let c_enmr = self.capi();
+        let mut c_str: *mut ffi::tiledb_string_t = out_ptr!();
+        self.capi_call(|ctx| unsafe {
+            ffi::tiledb_attribute_get_enumeration_name(ctx, c_enmr, &mut c_str)
+        })?;
+
+        if c_str.is_null() {
+            return Ok(None);
+        }
+
+        TDBString {
+            raw: RawTDBString::Owned(c_str),
+        }
+        .to_string()
+        .map(Some)
+    }
 }
 
 impl Debug for Attribute {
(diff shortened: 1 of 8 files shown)


tiledb/api/src/array/attribute/mod.rs
```diff
@@ -345,6 +365,25 @@ impl Builder {
         Ok(self)
     }
 
+    /// Set the name of the enumeration to use.
+    ///
+    /// Note that when building schemas, the enumeration must have been added
+    /// to the schema for adding an attribute that references it.
```
rroelke, line 371: Does failing to abide by this result in an error here?  Or when the attribute is added to the schema? Or at schema creation time?

tiledb/utils/src/option.rs
```diff
@@ -114,6 +114,24 @@ where
     }
 }
 
+impl<T> OptionSubset for Box<T>
```
rroelke, line 117: This probably oughta be more general, e.g.
`impl<T> OptionSubset for &T where T: OptionSubset`

I just had a look at `PartialEq` and it looks like it has both this version and also the one you added.  So, up to you if you want to add it.

tiledb/api/src/array/mod.rs
```diff
@@ -1238,4 +1266,79 @@ pub mod tests {
...
+            .build();
+
+            DomainBuilder::new(&ctx)?.add_dimension(dim)?.build()
+        };
+
+        let enmr = EnumerationBuilder::new(
+            &ctx,
+            "flintstones",
+            Datatype::StringUtf8,
+            "fredwilmageorgebetty".as_bytes(),
```
rroelke, line 1296: It's certainly not a request for changes in this PR but it feels a little wrong that the default cell val num is single but the other arguments here expect something else.

If offsets is `Some` then it should probably infer CellValNum::Var

rroelke (approved):
I'll leave a weak suggestion to add a test that the enumeration values which we put *in* to the schema are the same as those that we read *out*.  Otherwise this looks good

tiledb/api/src/array/attribute/mod.rs
davisp: You'll get an error from TileDB core yeah: https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/array_schema/array_schema.cc#L810-L820

tiledb/api/src/array/mod.rs
davisp: I'll add a story. It looks like Enumeration predates the CellValNum enum so it'll probably spider in the tests a bit. Also has the feel for another good "first story" for onboarding.

davisp: https://app.shortcut.com/tiledb-inc/story/54981

tiledb/utils/src/option.rs
davisp: Huh. I'll poke at that some more. I gave up a bit here when the first box wasn't doing exactly what I wanted and just added the `Box<[T]>` variant. Assuming your approach works I'll totes switch that out.

davisp: I attempted using this approach and couldn't quite figure out how to make it work correctly. The `impl` you suggested gets into a recursive loop that causes a stack overflow. Reading up more, I also found this post that seems to be related which just says that it has to do with fundamental types, which Option isn't.

PR #162, opened 2024-09-11
Move proptest strategy configurations into a single file

rroelke:
As requested in [this comment](https://github.com/TileDB-Inc/tiledb-rs/pull/157#pullrequestreview-2287147756), moving all the proptest strategy configuration to a single file is useful for maintainability/discoverability.

Note that the usage of `LazyLock` requires rust version 1.80.


davisp (approved):
+1

PR #166, opened 2024-09-24
chore: [sc-55233] [rs] panic in RecordsValueTree::simplify prevents proptest from dumping seed which led to bad input

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/55233

`RecordsValueTree` is an alternative to `proptest::collection::VecValueTree` for shrinking down a collection of values.

We have observed a few times in our tests which use this a panic during the shrinking process tracing to `RecordsValueTree`.  When we have such a panic, proptest shrinking iterations abort without printing the seed used to generate the failing input (or the current failing input).

`RecordsValueTree`'s shrinking logic was panicking in the expression `self.explore_results[i].take().unwrap()`.  This turned out to be a logical out-of-bounds access.  `i` was previously bounded by `self.explore_results.len()`, but not all of the indices would be filled in if there were fewer records than `explore_results`.

The fix is simple: bound `i` by the minimum of the number of records and the number of explorations.

# Testing

The test generates `RecordsValueTree`s together with arbitrary sequences of shrink steps (`ShrinkSequence`) and ensures that applying the arbitrary sequence does not result in a panic.

### DIFF
--- a/tiledb/test-utils/src/strategy/meta.rs
+++ b/tiledb/test-utils/src/strategy/meta.rs
@@ -1,0 +1,228 @@
+use std::fmt::{Debug, Formatter, Result as FmtResult};
+use std::rc::Rc;
+
+use proptest::prelude::*;
+use proptest::strategy::{NewTree, ValueTree};
+use proptest::test_runner::{Config as ProptestConfig, TestRunner};
+
+use crate::strategy::sequence::SequenceValueTree;
+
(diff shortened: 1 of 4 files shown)


davisp (approved):
You've... proptest'ed your proptest. So that's a thing now.

+1

Now with 100% more correct PR.

PR #169, opened 2024-09-26
chore: [sc-56223] [rs] implement custom proptest ValueTree for schema

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/56223

This pull request adds custom `proptest::strategy::ValueTree` implementations for `SchemaData` and friends.  This is beneficial for taking control of the shrinking algorithm but also allows us to construct a shrinkable ValueTree from a SchemaData, which will be very useful for debugging issues which originate outside of proptest.

The implementations ended up surprisingly simple: each `ThingValueTree` struct has fields which roughly correspond to `ThingData`.  `ValueTree::current` becomes trivial; `ValueTree::simplify` and `ValueTree::complicate` simply attempt to simplify or complicate fields in the order that we write them down.

Consider `SchemaValueTree`.  The first way we would want to shrink a schema is by identifying the set of necessary attributes.  `SchemaValueTree::simplify` and `complicate` begin by trying `selected_attributes.simplify` and `complicate` respectively.  This will converge the set of attributes, then move on to the next field to simplify/complicate.  This assumes that repeated calls to `simplify` and `complicate` are idempotent once the minimal set is found.  This property is required by `ValueTree`:

> The main invariants here are that the “high” value always corresponds to a failing test case, and that repeated calls to complicate() will return false only once the “current” value has returned to what it was before the last call to simplify().

although of course it needs proof for each `ValueTree` implementation.

`RecordsValueTree` abides this contract; `SequenceValueTree` does not and so this pull request updates it to do so.

### DIFF
--- a/tiledb/api/src/array/attribute/strategy.rs
+++ b/tiledb/api/src/array/attribute/strategy.rs
@@ -188,8 +188,8 @@ impl ValueTree for AttributeValueTree {
     fn current(&self) -> Self::Value {
         AttributeData {
             name: self.name.clone(),
-            datatype: self.datatype.clone(),
-            nullability: self.nullability.clone(),
+            datatype: self.datatype,
+            nullability: self.nullability,
             cell_val_num: self.cell_val_num.current(),
             fill: self.fill.current(),
             filters: self.filters.current(),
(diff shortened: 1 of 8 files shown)


davisp (approved):
+1

PR #170, opened 2024-09-27
feat: [sc-44563] Wrap Encryption related APIs

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/44563

This pull request enables interactions with encrypted arrays, and adds a unit test demonstrating creating an encrypted array and opening it, following along the C API example.

The encryption key is set on the tiledb Config object.  As this seems like a common use case, we add the convenience API `CommonOption` which makes configuration settings much more discoverable than writing the string in documentation somewhere.


tiledb/api/src/config.rs
```diff
@@ -265,6 +286,24 @@ impl<'cfg> Iterator for ConfigIterator<'cfg> {
     }
 }
 
+/// Convenience for setting some of the more commonly-used configuration options.
+#[derive(Clone, Debug)]
+pub enum CommonOption {
+    /// Sets an AES256GCM encryption key.
+    Aes256GcmEncryptionKey(Vec<u8>),
```
davisp, line 293: I wonder if this shouldn't be defined as something like:

```rust
pub enum CommonOption {
    Encryption(array::Encryption, Option<String>)
}
```

This is only a vague "might be better, might not" suggestion. So feel free to do what you want with it.

davisp (approved):
+1. The only thing I saw was the lack of setting "Unecrypted" via CommonOption, though I've not even convinced myself whether that would be useful or not so I'm not gonna hold things up over it.

tiledb/api/src/config.rs
rroelke: My thinking here (and the way I am following up in the tables repo so far) is that other encryption types might have more or different arguments.  So for now I'm just going with this.

PR #179, opened 2024-10-15
chore: [sc-57786] [rs] minimize set of functions which need to link with core tiledb to run

rroelke:
Story details: https://app.shortcut.com/tiledb-inc/story/57786

This pull request is a rather large "Part 1" of the linked story, whose goal is to restructure our crates in `tiledb-rs` so that the only code which requires linking with core tiledb is the code which actually calls the C APIs.  Doing this will simplify the build process for users of the `tiledb-rs` crates which do not need to use the C APIs directly, such as a client which sends RPCs to a server in tiledb-cloud.

We have written handles to tiledb schema objects: `struct Attribute`, `struct Dimension`, `struct Domain`, `struct Schema`, and so on.  When we turned our eye towards property-based testing, we concluded that we needed to write structures whose fields duplicate the fields of the objects inside of the core library (whereas the handles just provide methods to get or set them).  But over time we have come to use these secondary structures more and more, as they are often more user-friendly.  The goal of this "Part 1" pull request is to move these secondary structures to their own crate, which is tentatively named `tiledb-serde` in the diff.

Unfortunately this goal requires moving lots and lots of other pieces.

The new `tiledb` crates:

- `tiledb-common`: contains enums and other utility pieces, most notably including `Datatype`, `FilterData`, and `Range`.
- `tiledb-serde`: contains the secondary structures described above.  The name comes from the idea that these structures can be used to derive `Desesrialize` and `Serialize` for the API handles.  I think the name should be changed but I wanted to get things compiling and running again before taking that step.
- `tiledb-api`: renamed from the previous monolithic `tiledb`.

Each of the above crates has features `serde` and `proptest-strategies`.  The `serde` feature enables derived `Deserialize` and `Serialize` impls, but for `tiledb-api` this also adds `tiledb-serde` as a dependency.  The `proptest-strategies` feature similarly defines `Arbitrary` impls (for the most part) and is used as a `dev-depedency` by the `tiledb-api` crate.

There is a bit of `thiserror`-ization as entities which previously used `TileDBError` were moved into a new crate which does not have `TileDBError`.  But the `tiledb-api` crate itself still uses the large `TileDBError` enum.

Additionally, this pull request adds a new top-level directory to the workspace, `test-utils`.  This moves the various disjoint items from the previous `tiledb-test-utils` crate into their own crates.  All of the `Cells` testing code is also moved into a new crate `cells` in this directory.  This was mostly done to split out `strategy-ext` but the rest followed from there.

For the most part, code has just moved around, without any changes, but there are some exceptions:
- One difficulty with the crate splitting is our use of the `PhysicalType` trait.  Because this trait is now defined in a different crate than `tiledb-api`, we cannot provide generic impls on `PhysicalType` for some of the query stuff.  A new trait `CellValue` is introduced to work around this.
- Similarly the `cells` crate cannot have its `attach_read` and `attach_write` methods anymore since those use structures from `tiledb-api`, which uses on `cells` as a test dependency.  To resolve this the `tiledb-api` crate declares traits `ToReadQuery` and `ToWriteQuery` and provides impls for the items from `cells`.
- We previously used `FilterListData` to represent a filter list handle for proptests; this has been replaced with `Vec<FilterData>`.


test-utils/cells/Cargo.toml
rroelke, line 1: The `cells` crate contains the structure definitions and trait impls of code which used to live in `api/src/query/strategy.rs` and `api/src/query/write/strategy.rs`.  This code has been exported and is used in `tables` tests. This crate depends on `tiledb-serde` because it uses the schema to make some decisions about values and etc.

test-utils/proptest-config/Cargo.toml
rroelke, line 1: This new crate `proptest-config` just contains the configuration settings for proptest that we previously had.  I moved it to its own crate because `cells` and `tiledb-serde` both use it.

test-utils/signal/Cargo.toml
rroelke, line 1: We had some signal stuff in `tiledb-test-utils`, this crate contains that code.

test-utils/uri/Cargo.toml
rroelke, line 1: This crate is for the URI generator stuff which previously lived in `tiledb-test-utils`.

Initially I had split all of the `tiledb-test-utils` modules into separate *features* rather than separate crates; but this was sort of gnarly for dependencies, which is when I moved them all to their own crates.

tiledb/api/src/array/attribute/mod.rs
```diff
@@ -473,330 +476,18 @@ impl Builder {
     }
 }
 
-/// Trait for data which can be used as a fill value for an attribute.
-pub trait IntoFillValue {
```
rroelke, line 477: This was moved into `tiledb-common` because of the impls which build upon `PhysicalType`.   I did this before adding the `CellValue` trait; that would probably be more appropriate to use and we could move this trait back into `tiledb-api`.

tiledb/api/src/array/attribute/mod.rs
```diff
@@ -473,330 +476,18 @@ impl Builder {
...
-    }
-}
-
-impl Display for AttributeData {
-    fn fmt(&self, f: &mut Formatter) -> FmtResult {
-        write!(f, "{}", json!(*self))
-    }
-}
-
-impl TryFrom<&Attribute> for AttributeData {
```
rroelke, line 691: The conversions between `AttributeData` and `Attribute` now live in the `serde` module, i.e. `tiledb/api/src/array/attribute/serde.rs`.

This is the case for all the other `ThingData` structs as well

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -462,10 +191,30 @@ impl Builder {
         let constraints = constraints.into();
         constraints.verify_type_compatible(datatype)?;
 
-        let c_datatype = datatype.capi_enum();
+        let c_datatype = ffi::tiledb_datatype_t::from(datatype);
         let c_name = cstring!(name);
-        let c_domain = constraints.domain_ptr();
-        let c_extent = constraints.extent_ptr();
+        let c_domain = dimension_constraints_go!(
```
rroelke, line 196: This was the body of `DimensionConstraints::domain_ptr`.
It didn't really feel right to have an otherwise Rust-only crate provide methods which deal with raw pointers.

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -970,91 +525,28 @@ mod tests {
         }
     }
 
+    /// Test that the arbitrary dimension construction always succeeds
     #[test]
-    fn subarray_strategy_dense() {
```
rroelke, line 974: This test doesn't actually use any tiledb APIs so it now lives in `serde/src/array/dimension/strategy.rs`

tiledb/api/src/array/mod.rs
```diff
@@ -29,142 +28,21 @@ pub mod fragment_info;
...
 pub use fragment_info::{
     Builder as FragmentInfoBuilder, FragmentInfo, FragmentInfoList,
 };
 pub use schema::{
-    ArrayType, Builder as SchemaBuilder, CellValNum, Field, Schema, SchemaData,
+    ArrayType, Builder as SchemaBuilder, CellValNum, Field, Schema,
 };
-
-#[derive(Clone, Debug, PartialEq)]
-pub enum Mode {
```
rroelke, line 49: All the enums have been moved to `tiledb-common`

tiledb/api/src/datatype/arrow.rs
```diff
@@ -52,1033 +43,3 @@ impl ArrowPrimitiveTypeNative for f32 {
 impl ArrowPrimitiveTypeNative for f64 {
     type ArrowPrimitiveType = arrow::datatypes::Float64Type;
 }
-
-/// Represents tiledb (`Datatype`, `CellValNum`) compatibility for an arrow `DataType`.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub enum DatatypeToArrowResult {
```
rroelke, line 58: All of this is moved to `tiledb-common/src/datatype/arrow.rs`

tiledb/api/src/datatype/mod.rs
```diff
@@ -1,194 +1,34 @@
...
-use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+use crate::{FromStringCore, ToStringCore};
 
-use serde::{Deserialize, Serialize};
-use util::option::OptionSubset;
+pub use tiledb_common::datatype::*;
 
-use crate::error::DatatypeErrorKind;
-use crate::Result as TileDBResult;
+impl ToStringCore for Datatype {
```
rroelke, line 8: The enum is moved to `tiledb-common/src/datatype/mod.rs`.

What's here now is just trait impls for new traits `ToStringCore` and `FromStringCore` which call the tiledb core string functions.

Since these require linking with libtiledb, `tiledb-common` specifically does not want them.  Hence Debug is derived and Display just uses that.

tiledb/api/src/error.rs
```diff
@@ -26,107 +29,6 @@ impl Drop for RawError {
     }
 }
 
-#[derive(Clone, Debug)]
-pub enum DatatypeErrorKind {
```
rroelke, line 30: Some variants are moved to `tiledb_common::datatype::Error` and others are now just `Error` variants.

tiledb/api/src/filter/list.rs
```diff
@@ -183,75 +180,6 @@ impl Builder {
     }
 }
 
-#[derive(
-    Clone, Debug, Default, Deserialize, OptionSubset, PartialEq, Serialize,
-)]
-pub struct FilterListData(Vec<FilterData>);
```
rroelke, line 189: In contrast to other items which are moved somewhere else, this one I just deleted and replaced uses of it with `Vec<FilterData>`.

tiledb/api/src/filter/mod.rs
```diff
@@ -1,292 +1,25 @@
...
-mod ftype;
-mod option;
-
-use crate::filter::ftype::FilterType;
-use crate::filter::option::FilterOption;
-
-#[derive(
-    Copy, Clone, Debug, Deserialize, Eq, OptionSubset, PartialEq, Serialize,
-)]
-pub enum CompressionType {
```
rroelke, line 27: Moved to `tiledb_common::filter`

tiledb/api/src/metadata.rs
```diff
@@ -1,363 +1,44 @@
...
-use crate::physical_type_go;
-use crate::Result as TileDBResult;
-use core::slice;
-use std::convert::From;
-
-use serde::{Deserialize, Serialize};
-use util::option::OptionSubset;
-
-#[derive(Clone, Debug, Deserialize, OptionSubset, PartialEq, Serialize)]
-pub enum Value {
```
rroelke, line 12: This is moved to `tiledb_common`.  This module now just has functions for FFI.

tiledb/pod/src/array/attribute/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+/// Encapsulation of data needed to construct an Attribute's fill value
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "option-subset", derive(OptionSubset))]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub struct FillData {
+    pub data: MetadataValue,
+    pub nullability: Option<bool>,
+}
+
+#[cfg(any(test, feature = "proptest-strategies"))]
```
davisp, line 36: Is there any reason why you don't just move this impl to the strategy module?

davisp (commented):
Whoops. I totes forgot that I've got a dinner thing tonight so I didn't get as far into this as I'd like.

After my skimming so far, about the only thing I wish were different is the proptest strategies in the pod crate being pushed into their own. Granted I'm well aware of why that didn't and might not ever happen so its not really important. And its not like this isn't an improvement anyway.

Also heads up that there's a bunch of modules in the API crate all called serde that look like they should have been renamed to match the s/serde/pod/ rename you did earlier.

Other than that, this all looks pretty good. But given how large it is I am gonna give it another look over tomorrow before I approve. Today I mostly just looked at the new structure and spot checked that things pretty much ended up where I expected and so far. Tomorrow I'll give a closer skim on the actual diff before +1'ing.

Good work though! This will be a nice clean up of some of our initial decisions made while earlier in our Rust learning adventure.

tiledb/pod/src/array/attribute/mod.rs
rroelke: Just a stylistic choice to have the `impl` in the same file that the `struct` is defined in.
I have a weak preference for this because it means only searching one file for the impls. Though admittendly one should probably build rustdocs for exploring the impls.

tiledb/api/src/array/attribute/pod.rs
```diff
@@ -0,0 +1,105 @@
...
+            datatype,
+            nullability: Some(attr.is_nullable()?),
+            cell_val_num: Some(attr.cell_val_num()?),
+            fill: Some(fill),
+            filters: Vec::<FilterData>::try_from(&attr.filter_list()?)?,
+        })
+    }
+}
+
+impl TryFrom<Attribute> for AttributeData {
```
davisp, line 36: I assume this was just moved from wherever it was before, but seeing a try_from for both a ref and consuming seems odd. I can't quite convince myself this version would never be needed, but it sure seems unlikely.

Not asking for a change here. I'm just noting something I saw for future me writing trait implementations.

tiledb/api/src/array/dimension/mod.rs
```diff
@@ -521,216 +270,22 @@ impl From<Builder> for Dimension {
...
-    }
-}
-
 #[cfg(feature = "arrow")]
 pub mod arrow;
 
-#[cfg(any(test, feature = "proptest-strategies"))]
-pub mod strategy;
+#[cfg(any(test, feature = "pod"))]
+pub mod pod;
```
davisp, line 277: Still not sure on this pod as feature thing. Also, we should probably standardize that all module definitions (that aren't `tests`) should go at the top of the file. Finding the code/test transition isn't the easiest thing to navigate to just to check if there's anything interesting going on. Don't worry about that as part of this PR. Just another thing I've noticed while reading through.

tiledb/api/src/filter/ftype.rs
```diff
@@ -23,9 +39,56 @@ pub enum FilterType {
     Delta,
 }
 
-impl FilterType {
-    pub(crate) fn capi_enum(&self) -> ffi::tiledb_filter_type_t {
-        match *self {
+impl ToStringCore for FilterType {
```
davisp, line 42: I'm not entirely sure that `Core` is the best suffix here. Though I haven't got anything better to suggest so I can't really ask for a change...

tiledb/common/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dependencies]
+anyhow = { workspace = true }
+arrow-schema = { workspace = true, optional = true }
+paste = { workspace = true }
+proptest = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
+thiserror = { workspace = true }
+tiledb-proc-macro = { workspace = true, optional = true }
+tiledb-sys = { workspace = true }
```
davisp, line 16: I'm reading this PR pretty quickly, so I may have missed an actual ffi call, but I would only expect (and have only seen) references to various constants. We could pretty easily split those off into a tiledb-sys-defs crate to avoid this crate from requiring a link to libtiledb.so. Not needed for this PR, just a thought I had keeping in mind part of the motivation for this PR.

tiledb/common/src/filter/webp.rs
```diff
@@ -1,11 +1,19 @@
...
+pub enum WebPFilterError {
+    #[error("Invalid discriminant for {}: {0}", std::any::type_name::<WebPFilterInputFormat>())]
+    InvalidDiscriminant(u64),
+}
 
-#[derive(
-    Copy, Clone, Debug, Deserialize, Eq, OptionSubset, PartialEq, Serialize,
-)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "option-subset", derive(OptionSubset))]
```
davisp, line 15: Another feature that I'm not 100% sold on. Easy to change down the line, so not requesting a change. Just seems a bit "feature happy" is all.

tiledb/common/src/query/read.rs
```diff
@@ -0,0 +1,27 @@
+/*
```
davisp, line 1: Is this something that should be deleted? Looks like it was the start of something that was either deemed unnecessary or done differently.

tiledb/pod/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[dependencies]
+itertools = { workspace = true }
+num-traits = { workspace = true, optional = true }
+proptest = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
+strategy-ext = { workspace = true, optional = true }
+thiserror = { workspace = true }
+tiledb-common = { workspace = true }
```
davisp, line 15: Just a note to self that this hasn't removed the libtiledb.so linkage since we've got this transitive dependency. Though it looks pretty straightforward to cut that tie in a follow up PR that I assume is probably already half written waiting for me to approve this one. So on with the show!

tiledb/pod/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+//! "Plain old data" is used to describe types in C++ programming which
+//! have no constructors, destructors, or virtual member functions.
+//! Values of these types can be duplicated by copying bits.
+//!
+//! The structures defined in this crate are representations of tiledb
+//! logical structures which expose their attributes as public fields.
+//! This contrasts with the [tiledb-api] crate where the structures
+//! are instead handles to tiledb C API data structures.
+//!
+//! There is no direct "plain old data" analogue ini Rust; the use
```
davisp, line 12: s/ini/in/

tiledb/common/Cargo.toml
rroelke: Each time I add a new crate I have to add a new `build.rs` file for that crate.  Isn't that what sets the linker path, not the presence of the `sys` crate?

That said I do think it might be nice to split out the typedefs and constants for sys

tiledb/api/src/filter/ftype.rs
rroelke: I'm going back and forth on this one a bit.

When working within Rust only, the core "to string" function is a pain because the thing it prints out does not match the name of the enum variant.  This is a pain if you want to copy/paste code from a Debug trait output into code, which is something I have done a lot for converting proptest output into reproducers.

But the core to/from string functions are also the canonical representations which I assume are used by all the other language bindings.  So we want them for inter-op.

Maybe the answer is that Debug should do the Rust thing (i.e. `#[derive(Debug)]`) but we should use the core string functions for `Display` and `FromStr`?

Otherwise these traits aren't really used seriously so adding them to the list of ignored C API functions.  IIRC these are only present for Datatype and FilterOption.

tiledb/api/src/array/attribute/mod.rs
```diff
@@ -303,6 +291,19 @@ impl PartialEq<Attribute> for Attribute {
     }
 }
 
+#[cfg(any(test, feature = "pod"))]
```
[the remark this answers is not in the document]
rroelke: Well... it definiitely makes use of the API way easier but it's not *necessary* to use it.  If it's a feature than there is a leaner option to use tiledb as a dependency.

tiledb/api/src/array/attribute/pod.rs
rroelke: I don't really remember why I added this the first time, yeah.  Probably trying to do something mildly clever with iterators.

tiledb/common/Cargo.toml
davisp: There's two halves: The dependency on tiledb-sys introduces the requirement that libtiledb.so is linked to the final executable. The build.rs in non-tiledb-sys crates tells the linker where to find it on the current system (assuming its not in a system standard path).

So basically, if we split out tiledb-sys-defs that contains just the constant definitions, then anything that doesn't depend on tiledb-sys (directly or indirectly) will no longer need to provide the build.rs.

Or in other words, the build.rs requirement is a symptom of what's actually linking to libtiledb.so, when we remove that need we should be able to remove a lot of the build.rs scripts from the crates that no longer need it.

tiledb/api/src/array/dimension/mod.rs
rroelke: Yeah I think we have some stylization to do
- modules at the top
- `use` declaration un-grouping

tiledb/api/src/filter/ftype.rs
davisp: FWIW, I was only concerned about "core" vs something like "internal" since its what the library uses in a few places. It just felt like core being a word we use to refer to libtiledb.so and fiends might not carry the same meaning for someone not familiar with how we use that.

In terms of whether to even expose these, I don't know of any other bindings that use them to actually transfer meaning around vs "show me a debug description of this thing" like the dump functions for larger classes like Array and Query and so on. And given that we get that basically for free from Rust and `derive(Debug)`, I'd be fine just blanket ignoring all of the to_str/from_str functions right along with the dump functions.

tiledb/api/src/array/attribute/mod.rs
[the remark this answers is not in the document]
davisp: To me it was just that it seems like the pod definitions are. *the* way to interact with the bindings, so making it a feature that requires opting into didn't feel quite right. And this isn't super duper well thought out. I was just thinking "How often are we going to *not* want those pod APIs?" and I'm reasonably sure "Almost never, and the cost of having them and not using them would likely be miniscule."

Also its fine for now. We can always go back and change things up if we decide that it really is unnecessary for some particular use case where it matters.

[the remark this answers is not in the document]
rroelke: Indeed we could - and everyone loves PRs with lots of red

tiledb/api/src/filter/ftype.rs
rroelke: Great, I'll move them to ignored then!

tiledb/common/src/query/read.rs
rroelke: Derp. Deleleted

tiledb/pod/src/lib.rs
rroelke: Fixed

tiledb/common/Cargo.toml
rroelke: Ok I see that more clearly now from looking at `sys/build.rs`.  Definitely will follow up with that in a subsequent PR.  Thanks!

tiledb/api/src/filter/ftype.rs
rroelke: Removed the traits and impls and added the functions to `ignored.rs`

PR #182, opened 2024-10-21
Add libtiledb crate as a build dependency and do not use it in sys

rroelke:
The goal of this pull request is to ensure that the `common` and `pod` crates don't need to link with libtiledb.

[This discussion](https://github.com/TileDB-Inc/tiledb-rs/pull/179#discussion_r1809240293) suggested splitting the `sys` crate into a crate of types and constants, and a crate of functions.  This pull request approaches it differently.  The `sys` crate is really just a list of declarations - it is not itself runnable.  So we move the contents of its `build.rs`, which instruct cargo to link with libtiledb, over to the `api` crate instead, which is the lowest-level unit that actually needs to execute any linked code.

In doing so we move the build dependency code out of the `utils` crate and into a new crate named `libtiledb`.


davisp (changes requested):
No can do. Rust [naming conventions](https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages) are that the crate linking to `libfoo` should be named `foo-sys` and violating that for no good reason would be less than awesome. Obviously it can work as you've shown but it'd be awfully surprising to any Rust folks.

What we need is to add a tiledb-sys-defs (or whatever) crate and then just move `tiledb/sys/src/{constants.rs,capi_enums.rs,and maybe types.rs}` modules to them. You'll probably want to keep the existing modules and just do a public re-export of all symbols in the tiledb-sys-defs crate to avoid having to update a whole bunch of code.

That way common/pod/etc crates can just depend on the constants directly without a dependency on tiledb-sys.

We'll probably need to tweak the api coverage tool as well, but I think that should be easy enough because if memory serves I just recursively searched a directory so it shouldn't be hard to add a second. Originally I made that tool accept command line arguments, but if those are a pain to update we can just hard code things instead. I doubt anyone has ever once supplied an argument to that binary since I last tested the arguments work.

Also, for the build.rs stuff in crates that depend on tiledb-sys, I plan on adding a thing that'll make those much simpler. For now if you just do tiledb_sys::rpath() that'll be fine. In the future though I'll switch to using the [`links`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field) key with the `DEP_TILEDB_*` environment variables so we're only running pkg-config once.

rroelke:
Hey, my intuition is pretty good; I had a feeling you were going to push back on this way of doing it.

The rust naming convention is of course an excellent reason so I'll do a bit more reading about that and the `links` key especially.

I'll update this tomorrow with your requested changes.

(I was about to start with the `tiledb-sys-defs` approach and then realized I'd have to update the API coverage tool - this seemed like a path of lesser resistance. Shame it's wrong)

tiledb/queries/build.rs
```diff
@@ -1,5 +1,4 @@
 fn main() {
-    let libdir = pkg_config::get_variable("tiledb", "libdir")
-        .expect("Build-time TileDB library missing.");
-    tiledb_utils::build::set_linker_rpath(&libdir);
+    // NB: `api` crate dependency does `libtiledb::link()`
```
davisp, line 2: This comment is out of date.

davisp (approved):
Just had the one note on the out of date comment. Otherwise +1

PR #189, opened 2024-11-13
Robustness of attribute_enumeration_likelihood

rroelke:
`proptest::option::weighted` panics on inputs of 0.0 and 1.0.  This pull request adds checks for those values and provides an alternate strategy which implements the request.


davisp (approved):
+1 when CI is green

PR #190, opened 2024-11-13
impl conversions between RangeInclusive and SingleValueRange

rroelke:
`SingleValueRange` is essentially a dynamically-typed `std::ops::Range`.  This pull request implements conversions between the two.

I had this hanging around in a `git stash` and figured I'd clear it out.  I'm not sure what my original intent was but this seems potentially useful... or at least not harmful.


davisp (approved):
+1 Seems reasonable to me.

PR #191, opened 2024-11-20
`tiledb-rs` error styling part 1: `Context::capi_call` returns its own error type

rroelke:
In `tiledb-tables`, we've begun to declare at least one `Error` enum per module.  If a module `foo` calls a function from a module `bar`, then `foo::Error` has an enum variant approximating `Bar(bar::Error)`.  This can help build a causality chain in error output.

We have discussed bringing this error style into `tiledb-rs`, and this pull request is a foray into doing so.  We begin with the core of `tiledb-rs`, `Context::capi_call`.  This pull request adds `CApiError` which is returned by `Context::capi_call`.  The rest of the changes fell out of that in some way.

There are a few `// SAFETY` comments for new `unwrap`s.  These are added in situations where we try to convert a value returned from tiledb into one of the `tiledb-rs` enums.  The presupposition is that if libtiledb returned `TILEDB_OK` then it must have given a valid enum value.  I am personally comfortable with this assumption (I believe I am aligned with [this blog](https://blog.burntsushi.net/unwrap/#so-one-should-never-use-unwrap-or-expect) about `unwrap`) but I'm also fine changing it if requested - it would nudge the code in the direction of more `Error` variants.

### DIFF
--- a/tiledb/api/src/stats.rs
+++ b/tiledb/api/src/stats.rs
@@ -102,22 +102,13 @@ pub fn dump() -> Result<Option<String>, Error> {
 }
 
 #[cfg(feature = "serde")]
-mod __serde {
+pub fn dump_json() -> Result<Option<Vec<Metrics>>, Error> {
     use anyhow::anyhow;
-
-    use super::*;
-
-    #[cfg(feature = "serde")]
-    pub fn dump_json() -> Result<Option<Vec<Metrics>>, Error> {
-        if let Some(dump) = dump()? {
-            Ok(serde_json::from_str::<Vec<Metrics>>(dump.as_str())
-                .map(Some)
-                .map_err(|e| Error::ToJson(anyhow!(e)))?)
-        } else {
-            Ok(None)
-        }
+    if let Some(dump) = dump()? {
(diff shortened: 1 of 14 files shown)


tiledb/api/src/stats.rs
```diff
@@ -93,20 +102,22 @@ pub fn dump() -> TileDBResult<Option<String>> {
...
+                .map_err(|e| Error::ToJson(anyhow!(e)))?)
+        } else {
+            Ok(None)
+        }
     }
-    Ok(None)
 }
+
+#[cfg(feature = "serde")]
+pub use __serde::*;
```
davisp, line 123: I assume there's a good reason for doing this the way you've got it, but I'm not seeing it off the top of my head. Can you drop a comment explaining the setup/motivation?

davisp (approved):
Just had the one comment about the odd feature conditional compilation thinger.

tiledb/api/src/stats.rs
rroelke: I think this was really just to limit the number of `#[cfg(feature = "serde")]` but given that there isn't much in the module that seems a little silly.

PR #192, opened 2024-11-22
Add more `CommonOption`s

jparismorgan:
### What
Add more `CommonOption`s which we use when reading from `cloud.json` in TileDB-Tables. Part of [SC-54120](https://app.shortcut.com/tiledb-inc/story/54120/support-reading-cloud-config-api-tokens-from-tiledb-cloud-json).

### Testing
Adds new tests.

### DIFF
--- a/tiledb/api/src/config.rs
+++ b/tiledb/api/src/config.rs
@@ -294,6 +294,21 @@ impl Iterator for ConfigIterator<'_> {
 pub enum CommonOption {
     /// Sets an AES256GCM encryption key.
     Aes256GcmEncryptionKey(Vec<u8>),
+
+    // URL for REST server to use for remote arrays.
+    RestServerAddress(String),
+
+    /// Username for login to REST server.
+    RestUsername(String),
+
+    /// Password for login to REST server.
+    RestPassword(String),
+
+    /// Authentication token for REST server (used instead of username/password).
+    RestToken(String),
+
+    /// Have curl ignore ssl peer and host validation for REST server.
+    RestIgnoreSslValidation(bool),
 }
 
 impl CommonOption {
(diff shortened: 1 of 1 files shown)


tiledb/api/src/config.rs
```diff
@@ -294,6 +294,21 @@ impl Iterator for ConfigIterator<'_> {
 pub enum CommonOption {
     /// Sets an AES256GCM encryption key.
     Aes256GcmEncryptionKey(Vec<u8>),
+
+    // URL for REST server to use for remote arrays.
```
rroelke, line 298: Missing the third slash for a doc comment.

tiledb/api/src/config.rs
```diff
@@ -294,6 +294,21 @@ impl Iterator for ConfigIterator<'_> {
...
+    RestServerAddress(String),
+
+    /// Username for login to REST server.
+    RestUsername(String),
+
+    /// Password for login to REST server.
+    RestPassword(String),
+
+    /// Authentication token for REST server (used instead of username/password).
+    RestToken(String),
```
rroelke, line 308: Do we know if the token is always valid UTF-8?  If it isn't then `Vec<u8>` would be better

tiledb/api/src/config.rs
```diff
@@ -388,4 +421,48 @@ mod tests {
...
+            ("rest.username", "foo"),
+            ("rest.password", "bar"),
+            ("rest.token", "baz"),
+            ("rest.ignore_ssl_validation", "true"),
+        ];
+
+        for (key, val) in key_to_val {
+            let result: Option<String> =
+                cfg.get(key).expect("Error getting config key.");
+            println!("result: {:?}", result);
```
rroelke, line 458: Remove this before merging

rroelke (approved):
It might be worth checking about the UTF-8 question but the other comments are just minor tweaks. Thanks!

tiledb/api/src/config.rs
jparismorgan: Oops, thanks, done.

tiledb/api/src/config.rs
jparismorgan: Oops, thanks, done.

tiledb/api/src/config.rs
jparismorgan: Okay it is. From John Kerl: `it's always base64 so yes, & furthermore it's valid ASCII`.

PR #195, opened 2025-01-21
Rr/sc 44569 rs schema evolution

rroelke:
This pull request adds Rust APIs for schema evolution.

No proptests, but there is a unit test for each function except setting the timestamp range.

The C API doc says that the enumeration passed to `tiledb_array_schema_evolution_extend_enumeration` must have been returned from `tiledb_enumeration_extend`.  Using a newtype, we can enforce this requirement in the Rust API and we do so with the new `ExtendedEnumeration` type.


tiledb/api/src/array/schema/evolution/tests.rs
```diff
@@ -0,0 +1,260 @@
...
+        filters: Default::default(),
+        enumeration: Some(ename.to_owned()),
+    };
+
+    let states_enumeration = EnumerationData {
+        name: ename.to_owned(),
+        datatype: Datatype::StringAscii,
+        cell_val_num: Some(CellValNum::Var),
+        ordered: None,
+        data: "newhampshirenewjerseynewyork"
```
davisp, line 110: Why no love for New Mexico?

tiledb/sys-defs/src/lib.rs
```diff
@@ -193,3 +193,8 @@ pub type tiledb_vfs_mode_t = ::std::os::raw::c_uint;
 pub const tiledb_walk_order_t_TILEDB_PREORDER: tiledb_walk_order_t = 0;
 pub const tiledb_walk_order_t_TILEDB_POSTORDER: tiledb_walk_order_t = 1;
 pub type tiledb_walk_order_t = ::std::os::raw::c_uint;
+
+pub const tiledb_current_domain_type_t_TILEDB_NDRECTANGLE:
+    tiledb_current_domain_type_t = 0;
+#[doc = " The current domain type"]
```
davisp, line 199: Was there a reason to go the attribute route instead of `///`? I don't care which is used, I'm just curious if I'm missing something.

davisp (approved):
+1

tiledb/sys-defs/src/lib.rs
rroelke: Just a bad copy/paste

tiledb/api/src/array/schema/evolution/tests.rs
rroelke: Sorry new mexico :(

PR #196, opened 2025-01-21
First pass at proptest ValueTree for query condition

rroelke:
This pull request implements a custom `ValueTree` for query conditions.  This initial pass constructs a ValueTree given a query condition and shrinks the combination clauses.

Without a strategy to _generate_ a query condition, this can only be used to shrink to shrink one that has already identified, e.g.
```
use strategy_ext::StrategyExt;
Just(my_query_condition).value_tree_map(|vt| QueryConditionValueTree::new(vt.current())
```

The hope is that this can be useful for customer cases.  And someday when we build strategies to generate query conditions it can be useful for shrinking those, too.


davisp (approved):
+1

PR #199, opened 2025-02-17
Create nightly release packages

davisp:
Initial work on adding nightly releases to be consumed downstream.

The current plan is to switch over our existing CI pipelines to using pre-built packages of tiledb similar to how we're already handling this with the nightly-libtiledb releases. However, these will provide both nightly packages as well as static builds of upstream tags.

Doing it this way means that when I get to adding the tagged release pipeline the logic will all match other than extra bits about uploading the results to the version tag.

### DIFF
--- a/.github/scripts/generate-build-matrix.py
+++ b/.github/scripts/generate-build-matrix.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+
+# This is taken from the TileDB Release workflow found here:
+#
+# * https://github.com/TileDB-Inc/TileDB/blob/main/.github/workflows/release.yml
+#
+# The upstream version of this matrix does not include static builds which
+# can be useful for projects like TileDB-Tables that want to distribute
+# statically linked binaries.
+
+import copy
+import json
+
+BASE_MATRIX = [
+    {
+        "platform": "linux-x86_64",
+        "os": "ubuntu-20.04",
+        "manylinux": "quay.io/pypa/manylinux_2_28_x86_64",
+        "triplet": "x64-linux-release"
+    },
+    {
+        "platform": "linux-x86_64-noavx2",
+        "os": "ubuntu-20.04",
+        "cmake_args": "-DCOMPILER_SUPPORTS_AVX2=OFF",
+        "triplet": "x64-linux-release",
+        "manylinux": "quay.io/pypa/manylinux_2_28_x86_64"
+    },
+    {
+        "platform": "linux-aarch64",
+        "os": "linux-arm64-ubuntu24",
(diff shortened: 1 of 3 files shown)


davisp:
@rroelke A bit of background since this turned into a bit of a doozy in terms of CI updates.

The primary change in this PR is that I'm now building all of the packages we'll need in one central `Nightly TileDB Packages` workflow. This builds packages for everything in the [upstream release workflow matrix](https://github.com/TileDB-Inc/TileDB/blob/022af982bbc1c6b0587bf241ca5b882fce6113ee/.github/workflows/release.yml#L45-L72) except for packages that are available upstream. I.e., when we're using a tagged version of a shared libtiledb, we use packages from `TileDB-Inc/TileDB`. All other packages are consumed from a `nightlies` release on this repository.

With the newly built packages updated I then updated `tiledb-sys` and `tiledb-sys-cfg` to always rely on having pre-built libtiledb available so that it never attempts to build its own copy like I was doing for static builds.

With that out of the way, I then also updated the `Nightly CI` workflow to use the new packages and expanded matrix for checks.

And finally, this also updates the `Pull Request CI` workflow to use the new toys.

N.B., I'm about to update all references pointing at my fork to now point at this repository which will cause the CI to fail. Though I'll note that it passed PR CI before that update [here](https://github.com/TileDB-Inc/tiledb-rs/actions/runs/13423177086).

rroelke:
> All other packages are consumed from a nightlies release on this repository.

I think this answers my only question here, which is about how this affects us if we need to make a tweak in core prior to making changes to `tiledb-rs`.  My understanding of what you say here is that normal CI uses the latest nightly build of core.  If we need to add a new core API for use in tiledb-rs or whatever, then we would have to wait for the next nightly build whereas previously we would build master.

This seems like a fine trade-off so if I have it right then I'm happy to approve.  This looks like it was painful to put together, so thank you for your patience and expertise in doing so!

davisp:
> My understanding of what you say here is that normal CI uses the latest nightly build of core. If we need to add a new core API for use in tiledb-rs or whatever, then we would have to wait for the next nightly build whereas previously we would build master.

Correct that you'd have to wait for the `Nightly TileDB Packages` workflow to run, but with the caveat that you don't have to wait for it to kick off on schedule. I've highlighted the manual run button below if you've not seen the manual workflow trigger. Once it sees the sha changed on main, it'll rebuild all of the required packages and make them available for our CI. If memory serves, a full rebuild takes roughly 30m though it'll depend on whether any dependencies have changed since lots of things are cached.

Screenshot 2025-02-20 at 12 55 05 PM

PR #217, opened 2025-03-14
fix: correct results with query conditions on variable-length data SC-63464

rroelke:
This resolves [SC-63464](https://app.shortcut.com/tiledb-inc/story/63464/rs-incorrect-results-when-query-condition-on-variable-length-data).

`std::mem::size_of_val` was used with type inference.  The intent was deref coercion of the vector reference to `std::mem::size_of_val::<[T]>(&slice)` but actually ended up with `std::mem::size_of_val::<Vec<T>>(&vec)` which is always 24.

But this is a large PR, you say!  And you're right.  This bug demonstrated a gap in our test coverage.  In particular we have not actually run any queries with query conditions.  The rest of the contents of this PR is test code, or refactoring of non-test-code crates so to get the dependencies right for the test code:
- we move `enum QueryCondition` and such into `tiledb_common`, with traits in `tiledb_api` for building the C API condition
- we add a strategy for generating a lexicographic range between A and B (as a building block for generating query conditions)
- we add proptest strategies for generating query conditions
- we update our `test-utils/cells` crate to apply query conditions
- we add a property test which generates a random array and tests that we get the expected results back when applying a query condition


tiledb/api/src/query/strategy/mod.rs
```diff
@@ -0,0 +1,374 @@
+use std::collections::HashMap;
```
rroelke, line 1: I'm not sure why this is showing up as new, it was moved from `strategy.rs` to `strategy/mod.rs`

tiledb/api/src/query/strategy/tests.rs
```diff
@@ -0,0 +1,1012 @@
+use cells::write::strategy::{WriteParameters, WriteSequenceParameters};
```
rroelke, line 1: Everything in this file was split out of `strategy.rs` up until `fn instance_query_condition`

tiledb/common/src/query/condition/mod.rs
```diff
@@ -0,0 +1,930 @@
+#[cfg(feature = "proptest-strategies")]
```
rroelke, line 1: All the struct/enum definitions were moved from the `tiledb_api` crate so that I could use them in `test-utils/cells`.

There is some new stuff here, namely some of the `Not` impls and the `pub fn negate`

tiledb/api/src/query/strategy/mod.rs
davisp: Git's rename detection is all heuristics based and it gets things wrongs occasionally. No need to worry though because the rename detection doesn't actually do anything internally.

davisp (commented):
So I did a first quick skim and I'm left confused on where the actual fix is among all the other test related changes. Given the description I expect its rather small, but I couldn't spot it on a first pass. Mind giving me a tap with the ol' clue stick?

rroelke:
> So I did a first quick skim and I'm left confused on where the actual fix is among all the other test related changes. Given the description I expect its rather small, but I couldn't spot it on a first pass. Mind giving me a tap with the ol' clue stick?

Yeah that makes a lot of sense. It is quite buried.

https://github.com/TileDB-Inc/tiledb-rs/pull/217/commits/1b1565e6b23c0e30ae3a686aeb41b5bdde7554cd
https://github.com/TileDB-Inc/tiledb-rs/pull/217/commits/5b4d08f1dd0cec89a80ad526ce6728b78f172022

davisp (approved):
Ah! +1

PR #218, opened 2025-03-18
fix: physical type of `Datatype::Char` is architecture-dependent

rroelke:
The signed-ness of `char` is architecture-dependent.  We were previously using `i8` as the physical type.  Now that we are running query conditions this has manifested in test failures when comparing against negative values on architectures where `char` is actually unsigned.

This pull request fixes the issue by using `c_char` for the physical type of `Datatype::Char`.  `c_char` is just a [type alias](https://doc.rust-lang.org/std/ffi/type.c_char.html) for `u8` or `i8` depending on the architecture.

I also increased the shrinking bounds for nightly runs - we expect infrequent failures and when they do show up the smaller the better.  The changes here increase the number of shrinking iterations to a large number but bounds the shrink time at one hour.

### DIFF
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -55,6 +55,10 @@ jobs:
       - name: Build
         run: cargo build --all-targets --all-features
       - name: Test
+        env:
+          # give up shrinking after an hour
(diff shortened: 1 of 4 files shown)


davisp (approved):
+1

PR #220, opened 2025-03-25
fix: write query builder checks physical compatibility of buffers against field type

rroelke:
Resolves [SC-64904](https://app.shortcut.com/tiledb-inc/story/64904/rs-pr-ci-failing-for-platforms-when-std-ffi-c-char-is-unsigned).

This also fixes other `Datatype::Char` issues on arm64.


davisp (approved):
+1

PR #221, opened 2025-03-31
chore(rust): Upgrade to 1.85 and Edition 2024

davisp:
Mostly this is just removing a bunch of match modifiers along with a few other tweaks to make clippy happy.


rroelke (approved):
+1 !
Thanks!

PR #229, opened 2025-04-18
fix: Set the minimum fixed sized list to 2

davisp:
We caught a proptest failure because we translate a fixed size list of length 1, to a non-list version of the contained datatype. I.e., we changed the faux data type `[u32; 1]` to `u32`.


rroelke (approved):
This feels fine, I don't think we need to worry much about fixed-length list types of size zero or one

PR #230, opened 2025-04-25
chore: fix clippy::uninlined_format_args

rroelke:
This lint was turned on by default in nightly.  This PR cleans out the lint.

Enabling this lint by default strikes me as a bad decision.  As we can see in this PR, it increases the length of the format strings, which `cargo fmt` does not split onto multiple lines based on the desired display width; whereas it can do so more easily with arguments.

As an alternative we could disable this lint in our projects.

Since I already did the work here I'm fine either way; I don't mind the larger format strings too much, and it is *possible* to break them up on multiple lines - it just doesn't happen automatically.

### DIFF
--- a/test-utils/cells/src/lib.rs
+++ b/test-utils/cells/src/lib.rs
@@ -222,8 +222,7 @@ impl Cells {
         for k in keys.iter() {
             if !self.fields.contains_key(k) {
                 panic!(
-                    "Cannot construct view: key '{}' not found (fields are {:?})",
-                    k,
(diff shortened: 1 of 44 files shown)


davisp (approved):
+1

johnkerl (approved):
> Since I already did the work here I'm fine either way; I don't mind the larger format strings too much, and it is possible to break them up on multiple lines - it just doesn't happen automatically.

My $0.02 -- I can see arguments either way -- disable the lint, or move forward with this PR. My personal aesthetic is I do prefer the inlined version, FWIW. :)

PR #216, opened 2025-03-10
fix(lint): Fix new nightly lints

davisp:
This update addresses the `manual_contains` clippy lint.

PR #233, opened 2025-05-07
fix: Unbreak signal test; make CI test it

johnkerl:
#216 broke a signal test which fails for me (on bisect) with

```
cargo build --workspace --all-targets --all-features
cargo test --workspace --all-targets --all-features
```

This PR reverts that, and adds the `--workspace` flag to our CI YAML files. (Without that, only `default-members` in `Cargo.toml` are tested.)

### DIFF
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -53,13 +53,13 @@ jobs:
         # on that approach. So I'm just disabling the particular lint globally.
         run: cargo clippy --all-targets --all-features -- --allow non_local_definitions -Dwarnings
       - name: Build
-        run: cargo build --all-targets --all-features
+        run: cargo build --workspace --all-targets --all-features
       - name: Test
         env:
           # give up shrinking after an hour
           PROPTEST_MAX_SHRINK_ITERS: 100000000
           PROPTEST_MAX_SHRINK_TIME: 3600000
-        run: cargo test --all-targets --all-features
+        run: cargo test --workspace --all-targets --all-features
       - name: Check Linkage - Linux
         if: ${{ startsWith(matrix.os, 'ubuntu-') || startsWith(matrix.os, 'linux-') }}
         run: .github/scripts/ubuntu-check-${{ matrix.linkage }}-linkage.sh
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -38,10 +38,10 @@ jobs:
       - name: Install TileDB
         uses: ./.github/actions/install-tiledb
       - name: Build
-        run: cargo build --all-targets --all-features
+        run: cargo build --workspace --all-targets --all-features
       - name: Test
         run: |
-          cargo test --all-targets --all-features
+          cargo test --workspace --all-targets --all-features
           status=$?
           echo "Process exited with status ${status}"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,9 +33,9 @@ jobs:
       - name: Lint
         run: cargo clippy --all-targets --all-features -- -Dwarnings
       - name: Build
-        run: cargo build --all-targets --all-features
+        run: cargo build --workspace --all-targets --all-features
       - name: Test
-        run: cargo test --all-targets --all-features
+        run: cargo test --workspace --all-targets --all-features
 
   create_issue_on_fail:
     permissions:
--- a/test-utils/signal/src/lib.rs
+++ b/test-utils/signal/src/lib.rs
@@ -107,11 +107,10 @@ impl<'a> SignalCallback<'a> {
             restore_callback: None,
             action: Box::new(handler),
         });
-        let raw_callback = RawSignalCallback::from(callback.as_mut().get_mut());
-        SIGNAL_CALLBACKS[signo as i32 as usize]
-            .lock()
-            .unwrap()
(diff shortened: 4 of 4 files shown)


tiledb/common/Cargo.toml
```diff
@@ -22,6 +22,9 @@ tiledb-utils = { workspace = true, optional = true }
 proptest = { workspace = true }
 strategy-ext = { workspace = true }
 
+[build-dependencies]
+tiledb-sys-cfg = { workspace = true }
```
rroelke, line 26: I don't believe we need this here.  We should only need this dependency for crates which have `tiledb-api` somewhere in their dependency tree.

`tiledb-common` and `tiledb-pod` shouldn't need to link with libtiledb (even for `cargo test`) so my expectation is that they don't need this either.

.github/workflows/nightly-ci.yml
```diff
@@ -53,13 +53,13 @@ jobs:
         # on that approach. So I'm just disabling the particular lint globally.
         run: cargo clippy --all-targets --all-features -- --allow non_local_definitions -Dwarnings
       - name: Build
-        run: cargo build --all-targets --all-features
+        run: cargo build --workspace --all-targets --all-features
```
rroelke, line 56: I discovered (by running it) that this will not run all of the test binaries if a test fails in one of the crates.

We should consider using the [`--no-fail-fast` flag](https://doc.rust-lang.org/cargo/commands/cargo-test.html) as well:
> Run all tests regardless of failure. Without this flag, Cargo will exit after the first executable fails. The Rust test harness will run all tests within the executable to completion, this flag only applies to the executable as a whole.

Based on what I just ran in my checkout, this flag still has the process exit with non-zero if any of the tests failed - we might be able to just drop it in here?

rroelke: Oh, that's nice, it even lists the targets which failed:
```
error: 2 targets failed:
    `-p cells --doc`
    `-p tiledb-common --doc`
```
(for example)

tiledb/common/Cargo.toml
johnkerl: @rroelke we are here because I got errors (discussed in Slack) on MacOS CI:
https://github.com/TileDB-Inc/tiledb-rs/actions/runs/14888046535/job/41815828996
until I did these mods.

We may not need both `common` and `pod` but we need at least one.

I'll find out which one.

johnkerl: PRs #234 and #235 are experiments.

I'd do this on-laptop but I can't repro
https://github.com/TileDB-Inc/tiledb-rs/actions/runs/14888046535/job/41815828996
on-laptop so here we are with CI experiments.

rroelke: In that run I am seeing:

```
     Running unittests src/lib.rs (target/debug/deps/tiledb_common-ac48671f686ff452)

running 48 tests
...

     Running unittests src/lib.rs (target/debug/deps/tiledb_pod-d0ebd1d9a85f4f0c)

running 11 tests
...


     Running unittests src/lib.rs (target/debug/deps/tiledb_sys-02dc65d050e8e14f)
dyld[6094]: Library not loaded: @rpath/libtiledb.dylib
  Referenced from: <85A27323-041E-312F-82F5-692315DB31B9> /Users/runner/work/tiledb-rs/tiledb-rs/target/debug/deps/tiledb_sys-02dc65d050e8e14f
  Reason: tried: '/Users/runner/work/tiledb-rs/tiledb-rs/target/debug/deps/libtiledb.dylib' (no such file), '/Users/runner/work/tiledb-rs/tiledb-rs/target/debug/libtiledb.dylib' (no such file), '/Users/runner/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libtiledb.dylib' (no such file), '/Users/runner/.rustup/toolchains/stable-aarch64-apple-darwin/lib/libtiledb.dylib' (no such file), '/Users/runner/lib/libtiledb.dylib' (no such file), '/usr/local/lib/libtiledb.dylib' (no such file), '/usr/lib/libtiledb.dylib' (no such file, not in dyld cache)
error: test failed, to rerun pass `-p tiledb-sys --lib`
```

This looks to me like `tiledb-common` and `tiledb-pod` ran just fine, and then the link error happened with `tiledb-sys`.

rroelke: Since `tiledb-sys` doesn't have any tests (it just imports C symbols to Rust basically) maybe there's a way to skip it

johnkerl: For PR #234 CI fails with the `rpath` issue @davisp highlighted:
https://github.com/TileDB-Inc/tiledb-rs/actions/runs/14890053775/job/41820908128?pr=234

For PR #235 CI fails with the `rpath` issue @davisp highlighted:
https://github.com/TileDB-Inc/tiledb-rs/actions/runs/14890055312/job/41819645602?pr=235

So I think we need both `common` and `pod` -- @davisp , your thoughts?

johnkerl: > Since `tiledb-sys` doesn't have any tests (it just imports C symbols to Rust basically) maybe there's a way to skip it

See #238. Although this is non-optimal, if some developers need to remember to always skip `tiledb-ssy`. In our CI YAML we can hard-code whatever skip flags, of course.

johnkerl:
Replaced by #238

PR #236, opened 2025-05-07
test: Another experiment atop #233 (not for review)

johnkerl:
Another experiment atop #233 following 

https://github.com/TileDB-Inc/tiledb-rs/pull/233#discussion_r2078226838


johnkerl:
This didn't run CI. #237 did. Closing both.

PR #238, opened 2025-05-07
fix: Unbreak signal test; make CI test it

johnkerl:
We found that our CI was not running build/test with `--workspace`. As a result, some tests were not being run.

One of them was in the `signal` create, related to #216. On internal discussion, we decided that we are not using this crate and we don't want to keep it.

This PR removes that crate, and also adds the `--workspace` flag to our CI YAML files. (Without that, only `default-members` in `Cargo.toml` are tested.)

We also found that MacOS tests were failing (in CI only, not on-laptop, not on mine anyway) with

```
Running unittests src/lib.rs (target/debug/deps/tiledb_sys-02dc65d050e8e14f)
dyld[5493]: Library not loaded: @rpath/libtiledb.dylib
Referenced from: <85A27323-041E-312F-82F5-692315DB31B9> /Users/runner/work/tiledb-rs/tiledb-rs/target/debug/deps/tiledb_sys-02dc65d050e8e14f
Reason: tried: '/Users/runner/work/tiledb-rs/tiledb-rs/target/debug/deps/libtiledb.dylib' (no such file), '/Users/runner/work/tiledb-rs/tiledb-rs/target/debug/libtiledb.dylib' (no such file), '/Users/runner/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libtiledb.dylib' (no such file), '/Users/runner/.rustup/toolchains/stable-aarch64-apple-darwin/lib/libtiledb.dylib' (no such file), '/Users/runner/lib/libtiledb.dylib' (no such file), '/usr/local/lib/libtiledb.dylib' (no such file), '/usr/lib/libtiledb.dylib' (no such file, not in dyld cache)
error: test failed, to rerun pass `-p tiledb-sys --lib`
```

For this reason, we exclude `tiledb-sys`.


rroelke (approved):
Excellent, thank you!

johnkerl:
Thanks @rroelke !

PR #239, opened 2025-05-08
fix: BitsEq/BitsOrd/BitsHash impls for f32 and f64 to treat 0 == -0

rroelke:
We've recently seen the following appear in PR CI:
```
thread 'tests::query_condition_and' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/proptest-1.6.0/src/num/float_samplers.rs:462:1:
invalid range
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
but it turns out this has also failed in at least one nightly, which was reported in #231 .

The panic arises when `proptest-rs` asserts that the range it wants to sample is non-empty:
```
assert!(high - low > 0., "invalid range");
```
Breaking here with a debugger revealed that we have a range `-0..=0` which fails this assertion.

When does the `query_condition_and` test sample ranges? To construct atomic predicates for query conditions it samples the domain of its input (it does this so that the predicates are more likely to sometimes be true, e.g. `x = 12340765872934` matches `x` against random bits and is not likely to be true).  It follows that the strategy which does the sampling avoids constructing a range when the low and high bounds of the domain are equal; and something about -0 and 0 defeats that check.

Indeed that is exactly what we find in the following code:
```
                if lb.bits_eq(&ub) {
                    Just(Literal::from(lb)).boxed()
                } else {
                    (lb..=ub).prop_map(Literal::from).boxed()
                }
```
The `bits_eq` is the function of our custom trait `BitsEq`.  We use this trait (and sibling traits `BitsOrd` and `BitsHash`) rather than traditional float operations because those float operations are not reflexive (for equality) or not a total order (for comparison).  But the implementation of these misses the place where float comparison does the *opposite* thing, i.e. two distinct bit patterns comparing as the same logical value.

This pull request updates our `BitsEq`, `BitsOrd`, and `BitsHash` implementations so that the distinct bit patterns for 0 and -0 are treated equally.  We add tests for the logical properties requested of `Eq`, `Ord`, and `Hash` impls by the Rust standard library documentation.

### DIFF
--- a/tiledb/common/src/datatype/physical.rs
+++ b/tiledb/common/src/datatype/physical.rs
@@ -247,15 +247,34 @@ integral_type_impls!(i8: Datatype::Int8, i16: Datatype::Int16, i32: Datatype::In
 impl crate::private::Sealed for f32 {}
 impl crate::private::Sealed for f64 {}
 
+/// Defines an equivalence relation for `f32`.
+///
+/// The difference from the `PartialEq` implementation for `f32` is that of
+/// reflexivity. Specifically, if the bits match, then `self` and `other` are equal
+/// in this relation, whereas this is not true of floating-point equality for `NaN` values.
+///
+/// Zero and negative zero are considered equal under this relation.
 impl BitsEq for f32 {
     fn bits_eq(&self, other: &Self) -> bool {
-        self.to_bits() == other.to_bits()
+        self.to_bits() == other.to_bits() ||
+            // catch negative zero
+            (*self == 0f32 && *other == 0f32)
     }
 }
 
+/// Defines a total order for `f32`.
+///
+/// This ordering intends to be a bridge between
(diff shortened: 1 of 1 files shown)


johnkerl:
A minor (but persistent) annoyance re the one red CI bit on this PR (which is not this PR's fault!):

* Someone submits a PR with title "Lorem ipsum"
* PR-title CI runs & says "bad title"
* Submitter changes the title to "fix: Lorem ipsum"
* PR-title CI stubbornly refuses to let the grudge go
  * Example: https://github.com/TileDB-Inc/tiledb-rs/actions/runs/14896934430/job/41841181022?pr=239

After web-searching I think we can fix this in CI YAML with:
```
on:
  pull_request_target:
    types: [edited]
```
-- I'll look into that next week. It might be handy to split the PR-title munging to a separate CI YAML.

PR #242, opened 2025-05-14
feat: expose core group-exists API

johnkerl:
In the context of https://linear.app/tiledb/issue/TAB-57/misleading-error-on-read-of-tiledb-uris-in-tiledb-tables

### DIFF
--- a/tiledb/api/src/group/mod.rs
+++ b/tiledb/api/src/group/mod.rs
@@ -4,7 +4,7 @@ use crate::Datatype;
 use crate::Result as TileDBResult;
 use crate::config::{Config, RawConfig};
 use crate::context::{CApiError, Context};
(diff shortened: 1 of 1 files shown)


rroelke:
Even if it wasn't useful to solve the downstream problem this still looks useful

tiledb/api/src/group/mod.rs
```diff
@@ -77,6 +77,15 @@ impl Group {
         Ok(())
     }
 
+    pub fn exists<S>(context: &Context, uri: S) -> CApiResult<bool>
```
rroelke, line 80: I would just want to see at least one test for this

johnkerl: @rroelke definitely! This PR is in draft & not ready for review. I'll pick this up at some point (low-priority).

PR #245, opened 2025-05-29
chore(ci): Fix nightly lints

davisp:
Nightly has chainable if-let expressions which causes nightly clippy to trigger collapsible_if. Once chainable if-let is stabilized we can remove these and apply the collapse suggestion.


rroelke:
Seems like a good case for https://doc.rust-lang.org/reference/attributes/diagnostics.html `#[expect(C)]`

Except... would that happen only on nightly?

I'm approving anyway but thought I'd bring it up.

PR #259, opened 2025-12-10
chore: AttributeData and DimensionData can have parameterized name strategies in proptest

rroelke:
In `tiledb-tables` we will want to generate a schema with field names which are all lowercase. If we generate an arbitrary schema and then update the field names there is a small risk of collisions after lower-casing. Instead we can leverage the existing "no field name collisions" by enabling ourselves to only generate names which are already lowercase instead.


davisp (commented):
+1

PR #267, opened 2026-02-03
build: support Windows

teo-tsirpanis:
We use a fork of the pkg-config crate to skip passing `--cflags`, which we don't need, and causes private transitive dependencies to be searched.

Validated locally by building and running an example on Windows.

Fixes CORE-496.


tiledb/sys-cfg/build.rs
teo-tsirpanis: If libtiledb is obtained from the system, why do we need to override rpath?

